PR #84, opened 2026-02-12
Incorporate another round of wizard remarks

daniloc:
Next.js and PostHogProvider regularly confuse the agent, but the solution is clear: don't use it at all, the singleton is good to go thanks to `instrumentation-client.ts`. This enhances the commandments to clarify this point, with a reminder not to bother with edits until a file has been read (surprising source of friction!).

Also adds a changes explainer to the setup report, which will be especially helpful if the wizard trims out a PostHogProvider.


gewenyu99 (approved):
makes sense to me

PR #103, opened 2026-03-31
fix: route /array/* to assets origin

sarahxsanders:
related: https://github.com/PostHog/posthog.com/pull/16081

fixes /array/* path routing in context-mill example apps - was routing through the main API origin, which strips cache-control headers, causing browsers to cache stale SDK config for hours or days and breaking feature flags, session recording conditions, and sampling rates

also adds a javascript commandment so all client-side JS frameworks get the rule in one place. (note: javascript_web uses a separate tag so it gets its own copy)

### DIFF
--- a/basics/next-app-router/next.config.ts
+++ b/basics/next-app-router/next.config.ts
@@ -9,6 +9,10 @@ const nextConfig: NextConfig = {
         destination: "https://us-assets.i.posthog.com/static/:path*",
       },
       {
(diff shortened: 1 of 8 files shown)


edwinyjlim (approved):
totally missed this PR, sorry!

just one thing. we're hardcoding `us` into all the reverse proxy domains. I don't know if that'll affect the `eu folks. not a blocker

PR #108, opened 2026-04-13
feat: Best practices

gewenyu99:
Distilled the commandments and some small set of knowledge from docs here, really.

[posthog-best-practices.zip](https://github.com/user-attachments/files/26795627/posthog-best-practices.zip) (please ignore the fact that the proxy file here is in html, fixed later)

It's a little messy, I want to replace this ASAP with real best practice type checklists for each docs instead of having to do this manually, but it's a lean start for the greptile folks


edwinyjlim (approved):
so the combinatorial problem still crops up here `best practices skill = generic instructions x frameworks x products`. no easy solve for that, but it might be worth having at least one code snippet for each since we know that code examples improve quality by a lot. We can do that in the next iteration

gewenyu99:
> so the combinatorial problem still crops up here best practices skill = generic instructions x frameworks x products. no easy solve for that, but it might be worth having at least one code snippet for each since we know that code examples improve quality by a lot. We can do that in the next iteration

Yeah. I'm thinking we gotta resolve it with sourcing from docs where possible. This is just gonna be a huge headache if done here

PR #110, opened 2026-04-14
quack skill

edwinyjlim:
🦆 let's us test the new wizard harness with a lightweight skill (minimal detection and UX, only auth)

```
wizard --skill="quack-quack"
```


daniloc (approved):
:oh-yeah:

PR #112, opened 2026-04-15
Small tweak to remove placeholder

gewenyu99:
Thought about it, placeholder doesn't make sense. I think it should be able to find the distinct ID. I really don't think we're gonna have a situation where it can't with a valid posthog integration. 

This placeholder will just mask it with more confusion imo


sarahxsanders (approved):
agreed - should be able to find it

PR #114, opened 2026-04-15
no dots in skill names and paths

gewenyu99:
Rename the nuxt-3.6 skill id and directory to nuxt-3-6 so it passes the install_skill MCP tool's id validator (which rejects dots). 

Without this the agent falls back to a curl download of the skill zip, which fails with exit 7 from its sandboxed Bash and leaves it hand-rolling the Nuxt integration instead of using the authored skill.


sarahxsanders (approved):
good catch!

PR #115, opened 2026-04-15
feat: docs-as-skillz

sarahxsanders:
_docs as skills time_

auto generates a skill per section of the PostHog docs from `posthog.com/llms.txt`. each section becomes a downloadable skill with a SKILL.md + references/ directory. full explanation with changes in the README

## try it out

```
pnpm run build:docs-skills
cp -r dist/skills/posthog-feature-flags .claude/skills/
claude
> How do I bootstrap feature flags in PostHog?
```

it will load the skill and use the right file under the reference directory

related PR: 
- wizard PR (in progress)

follow up tasks: 
- posthog.com: add repository dispatch to deploy pipeline
- MCP PR: should we surface docs skills in the MCP?


.github/workflows/sync-docs-skills.yml
```diff
@@ -0,0 +1,184 @@
...
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
+        with:
+          version: 9
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Determine next version
```
gewenyu99: I feel like I've seen a bunch of versions of this same script. Can we reuse as an action or script or smthing?

.github/workflows/sync-docs-skills.yml
```diff
@@ -0,0 +1,184 @@
...
+
+      - name: Build docs skills
+        run: pnpm run build:docs-skills
+
+      # Compare only name fields in posthog docs entries (not install commands,
+      # which are stable but we want to catch additions/removals/renames).
+      - name: Check for changes
+        id: diff
+        env:
+          PREV_MENU_URL: ${{ github.server_url }}/${{ github.repository }}/releases/latest/download/docs-skill-menu.json
```
gewenyu99: This is probably fine, but asking these just in case:
- What if the contents of skills change but the menu doesn't
- And I guess, do we need to curl to a tmp file or can we use git? I think the docs-skill-menu.json is kept locally right (or maybe not...)

.github/workflows/sync-docs-skills.yml
```diff
@@ -0,0 +1,184 @@
+name: Sync PostHog Docs Skills
+
+# Triggered by posthog.com after a successful deploy, nightly as a fallback,
+# or manually. Fetches posthog.com/llms.txt, regenerates the posthog-docs
+# skill directories and docs-skill-menu.json, and cuts a new release if anything changed
+
+on:
+  repository_dispatch:
```
gewenyu99: We gotta debounce this somehow. I feel like this would end up being thousands of releases in a month or two :P

scripts/build-docs-skills.js
```diff
@@ -0,0 +1,344 @@
...
+
+// ---------------------------------------------------------------------------
+// HTTP helpers
+// ---------------------------------------------------------------------------
+
+/**
+ * Fetch a URL as text, retrying on failure.
+ * retries = 1 means a single attempt (no retries).
+ */
+async function fetchText(url, retries = 1, delayMs = 500) {
```
gewenyu99, line 49: This is gonna be huge. On every website build, these mds get physically generated. We should zip and upload these files as release artifacts on the website and download them to package (or smthing similar)

This is a lot of requests that we don't need that will take a loooonnnnggg time.

gewenyu99 (commented):
This is really cool!

I have two requests to look at:
- How do we avoid polluting the releases with way too many new releases. (we should have a maximum per day right?)
- Can we avoid requesting to the website, and just use the md files built during build time  for the posthog.com repo? It's all there, we don't need to download them from the website here.

scripts/build-docs-skills.js
```diff
@@ -0,0 +1,344 @@
...
+
+const LLMS_TXT_URL = 'https://posthog.com/llms.txt';
+const CATEGORY = 'posthog-docs';
+const CONCURRENCY = 10;
+const SKILLS_DIR = path.join(__dirname, '..', 'dist', 'skills');
+
+// Sections excluded by default — SDK and API reference material that is too
+// large and low signal-to-noise for skill context. Pass explicit slug args to
+// override and build one of these directly.
+const DEFAULT_EXCLUDE = new Set(['libraries', 'api', 'endpoints', 'open-api-spec']);
```
daniloc, line 39: smart

.github/workflows/sync-docs-skills.yml
```diff
@@ -0,0 +1,184 @@
+name: Sync PostHog Docs Skills
+
+# Triggered by posthog.com after a successful deploy, nightly as a fallback,
```
daniloc: YES

scripts/build-docs-skills.js
daniloc: > release artifacts

I like getting holistic about the build process in other repos, this is a good shout. much better to hit GitHub's storage for these mechanisms

.github/workflows/sync-docs-skills.yml
daniloc: is there a world where we replace the latest release instead of stacking it, at least when triggered from this path? I suppose that's bad hygiene in other contexts so may not be well supported...

daniloc:
> outputs go to docs-skill-menu.json which I intentionally kept seperate from skill-menu.json for now

I think that's wise! no sense swamping the wizard with all of these things

that said… I wonder if it would be useful to provide these in a secondary tool so the wizard could work out surprising problems on the fly based on docs. (I don't know if this is a good idea, the caffeine is still metabolizing. but it would certainly enhance the agency of the wizard)

scripts/build-docs-skills.js
```diff
@@ -0,0 +1,344 @@
...
+        // LLM has everything it needs in one place to pick the right file and cite it.
+        const referencesList = referenceFiles.length > 0
+            ? referenceFiles.map(r => {
+                const label = r.description ? `${r.title} – ${r.description}` : r.title;
+                const url = r.url.replace(/\.md$/, '');
+                return `- \`references/${r.filename}\` — ${label} (${url})`;
+            }).join('\n')
+            : null;
+        const bodyParts = [
+            `You are a PostHog documentation assistant. Use the content below to answer questions about ${section.heading}.`,
```
edwinyjlim: I know this is prob placeholder, but what do we think the use case is for encyclopedic skills with the entire product docs? What is the objective or codified process we want to help the dev accomplish here with a skill? 

low key, this feels like the hard part

Q&A is in the realm of InKeep, which can search across all of the docs, and people can connect and search with it via MCP

scripts/build-docs-skills.js
edwinyjlim: actually built quite fast for me, but yea we need to think about intermediate ways of storing and fetching zips

.github/workflows/sync-docs-skills.yml
edwinyjlim: agree we shouldn't cron our releases, would defeat the purpose of having versions if there's a bajillion of them

honestly I think it's okay to have explicit lag between website docs and context mill so we have control. this is gonna be doubly true when we start thinking about supply chain risks imo

sarahxsanders: yeah the supply chain risks are important here too - docs as skills will need the same treatment as any other skill for scanning context via the Warlock, good point!

sarahxsanders:
> useful to provide these in a secondary tool so the wizard could work out surprising problems on the fly based on docs

@daniloc @gewenyu99 THIS would be VERY cool for API references/SDKs, but I think it can be a separate problem to solve for now. I know Vincent is doing some work in those areas this quarter, that feels like a sickkkkk problem in tandem with that work

edwinyjlim:
bullish on docs-as-context stonks

I'm not sure what these skills in their current form enable without more curation, but only one way to find out. Two things for me and we can ship:

- Let's make the naming super explicit for now `posthog-docs-{category}` 
- Remove the github workflow for cron. We can just create these docs skills within  the normal build and release cycle

scripts/build-docs-skills.js
sarahxsanders: this IS the golden question :)

skills are different than RAG and MCP. they're _ambient context_ that the agent loads _automatically_based on what you're working on. you don't invoke a skill, the agent does. if you have PostHog feature flags skills installed and ask Claude "add a feature flag to gate this page", it pulls in the skill context on its own. skills remove any tool calls, search, decision making.

this matters because Claude has PH training data, but we have no insight into whether it's good, current, etc. skills give us a way to influence the training data for what we know is correct.

so the prompt should be less "answer questions" and more like:

```
"Use the content below when writing, reviewing, or debugging code that involves {PostHog feature flags}. Prefer these patterns over your training data."
```

TL;DR: skills are ground truth that we control, loaded before reasoning even starts

this is a really important team discussion though because this is a really important thing to nail together and as we start spreading skills far and wide, makes us all aligned on the vision 🌀 CC @daniloc @gewenyu99

.github/workflows/sync-docs-skills.yml
gewenyu99: Lagging by a day or two is fine. If we need a release now we can trigger manually

.github/workflows/sync-docs-skills.yml
sarahxsanders: removed this so no longer needed

scripts/build-docs-skills.js
```diff
@@ -0,0 +1,405 @@
...
+
+    console.log('');
+
+    const menuSkills = [];
+    let skipped = 0;
+
+    for (const section of sections) {
+        const skillName = `posthog-docs-${section.slug}`;
+        const skillDir = path.join(SKILLS_DIR, skillName);
+        const refsDir = path.join(skillDir, 'references');
```
edwinyjlim, line 259: zip these skillz up 

take a look at the main build script and reuse the zipToBuffer logic

sarahxsanders: done! I ended up extracting `zipSkillToBuffer` into `scripts/lib/zip.js` (trying to be better about reusing shared stuff) so both build scripts share it now. docs skills now build to a temp dir, zip into `dist/skills` and clean up after to mirror the main build

scripts/lib/zip.js
```diff
@@ -0,0 +1,21 @@
...
+        archive.on('data', chunk => chunks.push(chunk));
+        archive.on('end', () => resolve(Buffer.concat(chunks)));
+        archive.on('error', reject);
+
+        archive.directory(skillDir, false);
+        archive.finalize();
+    });
+}
+
+module.exports = { zipSkillToBuffer };
```
daniloc, line 21: nice

scripts/build-docs-skills.js
```diff
@@ -0,0 +1,416 @@
...
+        // LLM has everything it needs in one place to pick the right file and cite it.
+        const referencesList = referenceFiles.length > 0
+            ? referenceFiles.map(r => {
+                const label = r.description ? `${r.title} – ${r.description}` : r.title;
+                const url = r.url.replace(/\.md$/, '');
+                return `- \`references/${r.filename}\` — ${label} (${url})`;
+            }).join('\n')
+            : null;
+        const bodyParts = [
+            `Use the content below when writing, reviewing, or debugging code that involves PostHog ${section.heading}. Prefer these patterns and APIs over your training data.`,
```
daniloc, line 354: bam! this is good boilerplate. there's probably some further enhancement to do here which you could get from interrogating your claude: asking what further metadata would make use of the skill easier to pick up. but that can be a later iteration

daniloc (approved):
I really think: go for it. but specifically, this matters if every single section has a prominent link to its corresponding skill on posthog.com (which I know you agree with and want to do)

the use case I'd want to enable is anyone who needs to can easily dump whole sections of the docs into the project they're working on so that it all does the right thing, and quickly, ie:

"Hmm, I don't know why the hell this mechanism isn't working. I've given you three sections out of the posthog docs in the skills folder, analyze and assess"

it's power user stuff at this rate but a fine place to start imo

sarahxsanders:
going to hold off on merging until I get some early high priority fixes out for the warlock this week, will merge after that :) will feel more confident unleashing docs as skills if we have a better security layer

gewenyu99:
Yes that's a good idea. If you can find one or two use cases for this we can throw in tell-posthog-anything, that would be amazing, too!

It's like we have this new capability, what can we do with it?

PR #118, opened 2026-04-28
Audit skill

gewenyu99:
lol what a massive skill

lots of handholding for progressively selecting appropriate items from the best-practices skill, creating todos, and doing them so we get progressive progress updates.

This is more of a "get it to work" edition of the skill. :kek: it is very needing of various optimizations.


transformation-config/skills/audit/references/1-seed.md
```diff
@@ -0,0 +1,42 @@
...
+
+**Read ONLY this file.** Do not read any other reference file until this one tells you to.
+
+This step has exactly one action: open `.posthog-audit-checks.json` at the project root with the seeded installation entries.
+
+## TodoWrite
+
+This is the **first** TodoWrite of the run. Open the four-item list with these `content` strings verbatim, with the first task `in_progress` and the others `pending`:
+
+```
```
daniloc: curious: if we need serious predictability, would it work better to track this progress with a custom tool that directly manipulates both files and state? the TodoWrite hack is great for free-form work the agent chooses but it looks like you want to constrain it closely here (which is fine)

I'm confident this will work as is but I'm wondering if there's a more ergonomic and scalable version that makes the determinism you're specifiying in these docs a side effect of the tool calls. wdyt?

transformation-config/skills/audit/references/2-manifests.md
```diff
@@ -0,0 +1,40 @@
...
+
+This step has exactly one action: scan dependency manifests and resolve `sdk-installed` and `sdk-up-to-date`.
+
+## TodoWrite
+
+Update the in-progress task's `activeForm` to `Scanning dependency manifests`. Status and content stay the same.
+
+## Action
+
+1. `Glob` for the project's dependency manifests by name: `package.json`, `requirements.txt`, `pyproject.toml`, `Gemfile`, `composer.json`, `build.gradle`, `Podfile`. One Glob is enough. **Do not read any code**.
```
daniloc: note to self: I want to check if this procedure should also be a tool (feel free to argue against while I think it through @gewenyu99)

gewenyu99: I think it should be. It should in fact be static and something we can reuse.

transformation-config/skills/audit/references/1-seed.md
gewenyu99: Yeah we don't _need_ this. This is to make the already slow boot up feel faster so I don't stare at a blank table for like, longer than 2-3 minutes. I cannot for the life of me get this agent to putout a status update without first trying to make like 10 tool calls.

We can likely bypass this all together if we tried splitting this entire skill into subagent tasks.

daniloc: oh that's interesting, parallelize it with specific return value expectations, yeah

transformation-config/skills/audit/references/1-version.md
```diff
@@ -0,0 +1,75 @@
+# Step 1 — SDK installed + SDK up-to-date
+
+This step is intentionally narrow. It runs **before any other project work**. Resolve exactly two checks: `sdk-installed` and `sdk-up-to-date`. **Do not** read source code, locate init sites, look at `.env*` files, or scan for identify/capture call sites in this step — that all belongs to later steps.
+
+## TodoWrite
+
+This is the **first** TodoWrite of the run. Call `TodoWrite` with `todos` set to the **array** below (not a string — pass the literal array value):
```
daniloc: lol

this is a very specific form of cyberpunk

gewenyu99: There is a story behind every weird prompt.

transformation-config/skills/audit/description.md
```diff
@@ -6,9 +6,11 @@ Scope is intentionally narrow in this version: 9 core checks covering installati
 
 ## Workflow
 
-The audit runs as a 5-step chain. Each step file ends with a pointer to the next. The chain is split so the cheapest, highest-signal check (is the SDK installed and current?) resolves **first**, before any source-tree exploration.
+The audit runs as a 4-step chain. Each step file ends with a pointer to the next. Follow them in the order they are written. You must resolve them in order before any source-tree exploration.
 
-**Start by reading `references/1-seed.md`.** Do not Glob, ls, or find the skill directory. Do not preload future steps. Do not re-read a step file once you've moved past it. Do not re-read SKILL.md.
+The audit ledger is already seeded with the 9 pending checks. Use `mcp__wizard-tools__audit_resolve_checks` to patch each one as you finish it.
```
daniloc: is it still 9?

do we want some templating here or is that overkill?

gewenyu99: Still 9. Templating would be good. I'm gonna bring back the template. Had it at some point before pulling back to a smaller subset

scripts/lib/skill-generator.js
```diff
@@ -490,11 +491,43 @@ async function generateSkill({
...
+                const headingEnd = content.indexOf(headingMatch[0]) + headingMatch[0].length;
+                content = content.slice(0, headingEnd) + '\n\n' + refsConfig.preamble + content.slice(headingEnd);
+            }
+
+            // Auto-append continuation for sequential references
+            const nextFile = continuationMap.get(reference.name);
+            if (nextFile) {
+                content = content.replace(/\n+---\n+\*\*Upon completion, continue with:\*\*\s*\[.*?\]\(.*?\)\s*$/, '');
+                content += `\n\n---\n\n**Upon completion, continue with:** [${nextFile}](${nextFile})`;
+            }
```
edwinyjlim, line 510: My biggest suggestion: I think we should explicitly encode `preamble` and `sequential | next_step` patterns into frontmatter for each reference file

- relying on file naming convention is kind of clunky
- what happens if we need to eventually support a variety of files in `/references`, each with different preambles and sequences
- more composable if we do it on a per-file basis

.posthog-audit-checks.json
edwinyjlim, line 1: did you mean to check this in here? 

This could go into an `/assets` directory within the audit skill directory

gewenyu99: l o l no definitely not

scripts/lib/skill-generator.js
gewenyu99: agree

edwinyjlim (approved):

PR #119, opened 2026-04-28
Opt-in continuations and preambles at config.yaml

daniloc:
Probably won't be the last time we need this sort of structure, so let's give config.yaml the power to opt-in continuations and insert preambles.


gewenyu99 (approved):
fuuuggggg it's so hard to keep them on track. We'll be doing this a lot. Smaller files, more specific instructions are the only ways to keep it on track. 

We'll be doing this a lot for sure...

PR #124, opened 2026-05-01
feat: add warlock v1

sarahxsanders:
merge after this one is shipped https://github.com/PostHog/warlock/pull/27

the warlock is here. wiring this up as a git dependency for now, will swap to the npm package once its published

## changes

- created `scan-warlock.js` - Node.js replacement for `scan-prompt-injection.sh`
- added `pnpm scan:skills` to scan built ZIPs and `pnpm scan <file>` to scan individual files locally
- updated CI workflows to use the new script
- added `pnpm-workspace.yaml` to allow warlock's build scipt to run on install
- added llm triage layer >:]
- added PH tracking for results https://us.posthog.com/project/2/dashboard/1655489

## known limitations, going to address in a follow up

- git dependencies are slower than npm, this is just temp until we publish to npm. it's an easy swap

## test

```
pnpm install
pnpm build
pnpm scan:skills                    # scan all built skill ZIPs
pnpm scan path/to/some/file.md      # scan a specific file
```


package.json
```diff
@@ -10,13 +10,17 @@
     "test:plugins:watch": "vitest scripts/plugins/tests",
     "test:skills": "vitest run scripts/lib/tests",
     "test:skills:watch": "vitest scripts/lib/tests",
-    "test": "vitest run scripts/plugins/tests scripts/lib/tests"
+    "test": "vitest run scripts/plugins/tests scripts/lib/tests",
+    "scan": "node scripts/scan-warlock.js",
+    "scan:skills": "node scripts/scan-warlock.js dist/skills"
```
edwinyjlim: let's rename to warlock or warlock-scan or security-scan, something that implies safety check

scripts/scan-warlock.js
```diff
@@ -0,0 +1,334 @@
...
+      if (!uniqueFiles.has(label)) {
+        uniqueFiles.set(label, content);
+      }
+    }
+    const combinedContent = [...uniqueFiles.entries()]
+      .map(([label, content]) => `--- ${label} ---\n${content.slice(0, 2000)}`)
+      .join("\n\n");
+
+    const triaged = llmProvider
+      ? await triageMatches(combinedContent, rawMatches, llmProvider)
```
edwinyjlim, line 382: might need to handle the scenario where the message becomes too large

edwinyjlim (approved):
STOKED

scripts/scan-warlock.js
sarahxsanders: maybe a 100k combined cap? haiku's context is 200k so it gives us room to bump it up if needed

edwinyjlim: I was thinking more on the lines of creating the conversation loop. say the combined content is 1M tokens, we should iterate over them in whatever batch size

sarahxsanders: OHHHHHH yes that make sense

sarahxsanders: added batching for matches! tested locally and a full context-mill scan has two batches. also this makes sure nothing ever gets skipped, so great catch :D

sarahxsanders: also that change is in the warlock to keep the clear warlock vs. consumer boundary

edwinyjlim: NOICE

.github/workflows/build.yml
```diff
@@ -27,14 +27,32 @@ jobs:
...
+      uses: actions/create-github-app-token@d72941d797fd3113feb6b93fd0dec494b13a2547 # v1.12.0
+      with:
+        app-id: ${{ secrets.GH_APP_POSTHOG_WIZARD_CI_BOT_APP_ID }}
+        private-key: ${{ secrets.GH_APP_POSTHOG_WIZARD_CI_BOT_PRIVATE_KEY }}
+        repositories: warlock
+
+    - name: Configure git auth for private deps
+      env:
+        APP_TOKEN: ${{ steps.app-token.outputs.token }}
+      run: git config --global url."https://x-access-token:${APP_TOKEN}@github.com/".insteadOf "https://github.com/"
```
gewenyu99, line 44: Not a lot of duplication, and not something we have to do now, but it does seem a good chunk of the build and build-release workflows can be turned into an action and reused

scripts/scan-warlock.js
```diff
@@ -0,0 +1,415 @@
...
+  ".md",
+  ".txt",
+  ".yaml",
+  ".yml",
+  ".json",
+  ".js",
+  ".ts",
+  ".py",
+  ".rb",
+  ".sh",
```
gewenyu99, line 38: Is it intended that files like `.swift` are omitted here? Is this supposed to be the complete set?

scripts/scan-warlock.js
```diff
@@ -0,0 +1,415 @@
...
+  if (!apiKey) return null;
+
+  const client = new Anthropic({
+    baseURL,
+    apiKey,
+  });
+
+  return async (prompt) => {
+    const res = await client.messages.create({
+      model: "claude-haiku-4-5-20251001",
```
gewenyu99: Nit: lift this to be a var instead of hard coding. This will probably reappear in places and change somewhat frequently

gewenyu99 (approved):
1 other thing, can you make sure relevant skills and .md files are updated for agents to be unconfused?

scripts/scan-warlock.js
sarahxsanders: I've expanded to make it a more complete set!

.github/workflows/build.yml
sarahxsanders: will defer to follow up PR!

PR #129, opened 2026-05-11
Events audit skill

edwinyjlim:
Skill to audit, analyze, and then create a report of your codebase's PostHog event map.

1. Detect SDK
2. Grep for event captures
3. Enrich events with analysis and metadata
4. Query MCP for event volume data
5. Create events audit markdown report
6. Create events audit dashboard

image

image

image


sachinr:
Got a customer who is looking for this + the ability to detect places in the code that aren't emitting events that they should (their example is that they have multiple code paths that result in images being generated but only some of them emit `image_generated`). Do you think that that should be part of this skill or a separate one?

edwinyjlim:
@sachinr is this what you need? 

Claude came up with buzz term "phantom events" but I think it's exactly what you're describing: instrumented events in the codebase that have no actual volume in PostHog or don't fire.

image

sachinr:
Slightly different - they want call sites in their code that _should_ emit events but don't currently. They're basically looking for a list of events in PostHog that aren't actually comprehensively capturing all user interactions.

transformation-config/skills/events-audit/config.yaml
```diff
@@ -0,0 +1,14 @@
+type: docs-only
+template: description.md
+description: Audit PostHog events in a codebase — produce an inventory of every captured event mapped to its file, area, and 30-day volume for the product team to query
+tags: [analytics, audit, best-practices]
+references:
+  preamble: "**Read ONLY this file.** Do not read any other reference file until this one tells you to."
```
daniloc, line 6: nice

daniloc (approved):
this is probably the most complex staged thing we have now; worth giving some thought to how we keep it tested?

PR #137, opened 2026-05-20
Add Django and Flask feature skill configs

marandaneto:
## Problem

Django and Flask apps only had generic Python skill coverage for feature flags and error tracking, even though their framework docs include implementation details that agents need. The audit events optimization reference also used a `$host` property in an example query, which triggered the security scanner's request-host rule.

## Changes

- Added Django and Flask variants to the feature flags skill config.
- Added Django and Flask variants to the error tracking skill config.
- Added Django and Flask library docs to the omnibus feature flags and error tracking skill configs.
- Updated the audit events environment-pollution query to use `$current_url` instead of `$host` and adjusted the guidance to inspect hostnames from URLs.
- Verified the updated skill configs build successfully.

## Testing

- [x] `pnpm test:skills`
- [x] `pnpm build`

### DIFF
--- a/transformation-config/skills/audit-events/references/3-events-optimize.md
+++ b/transformation-config/skills/audit-events/references/3-events-optimize.md
@@ -159,19 +159,19 @@ Procedure:
 ```sql
 SELECT
   properties.$lib AS lib,
-  properties.$host AS host,
+  properties.$current_url AS current_url,
   properties.$app_version AS app_version,
   count() AS n
 FROM events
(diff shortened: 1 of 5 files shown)


marandaneto:
im not sure if integrations should be added to the standard feat config files, but if not, happy to close it

edwinyjlim (approved):
@marandaneto tbh we don't know which combination of context in the skills references is optimal. only way to know is to try it out!

PR #140, opened 2026-05-22
only reload and rebuild skills that are being edited

edwinyjlim:
Before: every edit rebuilds the entire context mill library

After: edits trigger only partial rebuilds scoped to the specific skill group

local-rebuilt

### DIFF
--- a/scripts/dev-server.js
+++ b/scripts/dev-server.js
@@ -3,171 +3,282 @@
 /**
  * Development server for MCP resources
  *
- * Serves the generated ZIP file over HTTP and watches markdown files
- * for changes, automatically rebuilding when needed.
+ * Runs one full build at startup, then watches transformation-config/skills/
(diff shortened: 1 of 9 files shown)


scripts/dev-server.js
```diff
@@ -3,227 +3,335 @@
...
+    if (removed.length > 0) {
+        console.log(`🗑  Reconciled ${removed.length} orphan ZIP(s) from prior runs`);
+    }
 
-    // Start server
     createServer();
-
-    // Setup file watchers
-    setupWatchers();
+    setupWatcher();
```
gewenyu99: Do we need to call close() on the watcher to clean it up or does it just stop when this process is killed?

gewenyu99 (approved):
noice

scripts/dev-server.js
edwinyjlim: you're so RIGHT

PR #145, opened 2026-05-26
Add .NET skill coverage

marandaneto:
## Problem

The PostHog skill resources had a .NET feature flag variant, but did not cover the rest of the .NET SDK surface: initial integration, product analytics instrumentation, ASP.NET Core Feature Management, error tracking, or PostHog.AI/.NET LLM analytics.

## Changes

- Add minimal `basics/dotnet` and `basics/aspnetcore` example projects.
- Add `.NET` and `ASP.NET Core` integration variants.
- Add .NET examples/docs to omnibus integration and product analytics skills.
- Add .NET / ASP.NET Core error tracking references.
- Add an ASP.NET Core Feature Management-specific feature flag variant.
- Add PostHog.AI / .NET OpenAI coverage to LLM analytics skills.
- Add `dotnet` and `aspnetcore` commandments.
- Ignore .NET `bin/` and `obj/` outputs in repo and example processing.
- Align sample docs with the actual environment variable fallbacks used by the examples.

## Testing

- `dotnet build basics/dotnet/PostHogDotNetExample.csproj`
- `dotnet build basics/aspnetcore/PostHogAspNetCoreExample.csproj`
- `pnpm test`
- `pnpm build`


transformation-config/skills/llm-analytics/config.yaml
```diff
@@ -2,19 +2,21 @@
...
   - id: setup
     display_name: all supported providers
-    tags: []
+    tags: [python, javascript]
     docs_urls:
       # --- Model providers ---
       - https://posthog.com/docs/llm-analytics/installation/openai.md
       - https://posthog.com/docs/llm-analytics/installation/azure-openai.md
+      # --- PostHog.AI / .NET OpenAI instrumentation ---
+      - https://raw.githubusercontent.com/PostHog/posthog-dotnet/main/src/PostHog.AI/README.md
```
marandaneto, line 19: This is not documented in the posthog.com/docs, soIi figured linking like this is ok? until we get this documented

basics/aspnetcore/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+        options.PersonalApiKey ??= builder.Configuration["POSTHOG_PERSONAL_API_KEY"];
+
+        if (Uri.TryCreate(builder.Configuration["POSTHOG_HOST"], UriKind.Absolute, out var hostUrl))
+        {
+            options.HostUrl = hostUrl;
+        }
+    });
+
+    // Enables Microsoft.FeatureManagement (`IFeatureManager`, feature tag helpers,
+    // and FeatureGateAttribute) backed by PostHog feature flags.
```
dustinbyrne: nit: `FeatureGateAttribute` and tag helpers require `Microsoft.FeatureManagement.AspNetCore`, which is not demonstrated (or enabled) here

basics/dotnet/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+        personPropertiesToSet: new Dictionary<string, object>
+        {
+            ["email"] = "dotnet-demo@example.com",
+            ["runtime"] = ".NET",
+        },
+        personPropertiesToSetOnce: new Dictionary<string, object>
+        {
+            ["first_seen_in"] = "dotnet-console-example",
+        },
+        CancellationToken.None);
```
dustinbyrne: worth demonstrating `EvaluateFlagsAsync`?

dustinbyrne (approved):
looks good, I left a couple of comments/recommendations but up to you what you'd like to do

basics/aspnetcore/.env.example
```diff
@@ -0,0 +1,4 @@
+POSTHOG_PROJECT_TOKEN=phc_your_project_token_here
+POSTHOG_HOST=https://us.i.posthog.com
```
gewenyu99:
```suggestion
POSTHOG_HOST=https://us.i.posthog.com
# POSTHOG_HOST=https://eu.i.posthog.com
```
Save a potential round trip for the agent

basics/aspnetcore/Program.cs
```diff
@@ -0,0 +1,123 @@
...
+    }
+});
+
+app.MapGet("/", () => "PostHog ASP.NET Core example");
+
+app.MapPost("/api/todos", async (TodoRequest request, HttpContext context, IPostHogClient posthog) =>
+{
+    var distinctId = GetDistinctId(context);
+
+    await posthog.IdentifyAsync(
```
gewenyu99: These apps are very simple, which is great. What's not great is that context dependent calls, like identify, doesn't have an inferred shape.

The point of examples in this repo aren't to provide a shape the agents can follow.

I think the reason examples are so helpful, is to help the agent piece together that:
- I'm a server side sdk
- I receive the distinct Ids from the client side, which has called identify to create a stable Id.
- I will use the same sessionId for session continuity.

The way it's used here might be more confusing than helpful for agents. I think this is a pretty unnatural place to call identify. 

Maybe if the API reveals something new about the person, like they updated their name or email, we should be calling `identify` like this with `personPropertiesToSet()`

transformation-config/skills/llm-analytics/config.yaml
gewenyu99: 🤔 If it builds it's fine! Make sure this doesn't violate their open source license, somehow, tho.

gewenyu99 (commented):
Hey, this is amazing! Really grateful that you're helping out with these skills.

A general comment before we ship this. The examples being simple is the correct way to go, but think carefully about the inferred shape of the example.

They're really to provide the LLM a fuzzy shape of what types of events should be captured, where in the life cycle should we be identifying, what should be done with a session ID, etc.

I think these example demonstrate the _shape_ of the API, which is great, but the docs can do this, too. I would recommend you also look into how this example helps the LLM infer the place in life cycle that these code snippets should be called. Something slightly more realistic would be enough.

transformation-config/skills/llm-analytics/config.yaml
marandaneto: > Make sure this doesn't violate their open source license, somehow, tho.

not sure i follow? this is the readme from our own repo

basics/aspnetcore/Program.cs
marandaneto: Good point — fixed in a7017e2 by narrowing the sample comment to only mention `IFeatureManager`, which is what this example actually demonstrates.

marandaneto:
@gewenyu99 that makes sense, i thought we just needed a very simple example, what i would do is to remove the basic example from this pr to unblock the pr and follow up next, cus i have to do this for some other sdks and it'd be much more work than planned (ps i learned last week about the basic examples)
so i will keep adding the missing docs/configs/entries/without the sample for now and come back to it later

marandaneto:
> looks good, I left a couple of comments/recommendations but up to you what you'd like to do

removed the sample for now because of https://github.com/PostHog/context-mill/pull/145#issuecomment-4553128453

gewenyu99:
All good. This makes a lot of sense. The examples are useful when good, but this will work without examples, too.

PR #152, opened 2026-05-27
Use captureException SDK methods in error tracking examples

gewenyu99:
## Summary

Manoel flagged in Slack that some of our examples were capturing `$exception` events through the generic `capture` method with hand-built property bags. This PR switches each example to the SDK's first-class exception capture API so PostHog receives proper exception metadata (correct `$exception_list` shape, fingerprinting, etc).

## Changes per example

- **Angular** (`posthog-js`): use `posthog.captureException(error)`
- **Expo** & **React Native** (`posthog-react-native@4`): use `posthog.captureException(error, additionalProperties)`
- **Ruby** (`posthog-ruby`): use `posthog.capture_exception(e, distinct_id)`; bump `Gemfile` to `~> 3.3` (first 3.x to ship `capture_exception`)
- **Laravel** (`posthog-php`): delegate `PostHogService::captureException` to `PostHog::captureException`. The previous wrapper built properties without the `$` prefix (`exception_type`, `exception_message`, ...) so PostHog never parsed them as exception fields. Bumps `composer.json` to `^4.2` and updates `composer.lock` to 4.2.0 (first 4.x with `PostHog::captureException`).
- Updated each example README snippet to match.

## Test plan

- [x] `ng build` succeeds for the Angular example with the new code
- [x] `tsc --noEmit -p .` on Expo example reports no new errors (only pre-existing unrelated errors)
- [x] `tsc --noEmit -p .` on React Native example reports no new errors in `ProfileScreen.tsx`
- [x] Run Laravel + Ruby examples end-to-end against a real PostHog project (requires PHP/Ruby toolchains — not available in this sandbox; needs manual smoke test)
- [x] Confirm `$exception` events show up in PostHog Error Tracking with parsed stack frames

Tested these locally to make sure they work
Screenshot 2026-05-27 at 12 09 27 PM

---

### DIFF
--- a/basics/expo/app/profile.tsx
+++ b/basics/expo/app/profile.tsx
@@ -32,7 +32,7 @@ export default function ProfileScreen() {
   /**
    * Triggers a test error and captures it in PostHog
    *
-   * This demonstrates manual exception capture using the $exception event.
+   * This demonstrates manual exception capture via captureException.
    * In production, you would typically set up automatic exception capture
    * or use the before_send callback for customization.
    *
@@ -44,21 +44,9 @@ export default function ProfileScreen() {
     } catch (err) {
       const error = err as Error
 
-      // Capture exception in PostHog
       // @see https://posthog.com/docs/error-tracking
-      posthog.capture('$exception', {
-        $exception_list: [
-          {
-            type: error.name,
-            value: error.message,
-            stacktrace: {
-              type: 'raw',
-              frames: error.stack ?? '',
-            },
-          },
-        ],
+      posthog.captureException(error, {
         $exception_source: 'react-native',
-        // Additional context
         username: user.username,
         screen: 'Profile',
       })
--- a/basics/expo/app/profile.tsx
+++ b/basics/expo/app/profile.tsx
@@ -46,7 +46,6 @@ export default function ProfileScreen() {
 
       // @see https://posthog.com/docs/error-tracking
       posthog.captureException(error, {
-        $exception_source: 'react-native',
         username: user.username,
         screen: 'Profile',
       })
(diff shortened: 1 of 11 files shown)


marandaneto:
i think @PostHog/team-error-tracking would know better

basics/expo/app/profile.tsx
```diff
@@ -44,21 +44,9 @@ export default function ProfileScreen() {
...
-            type: error.name,
-            value: error.message,
-            stacktrace: {
-              type: 'raw',
-              frames: error.stack ?? '',
-            },
-          },
-        ],
+      posthog.captureException(error, {
         $exception_source: 'react-native',
```
hpouillot: $exception_source is not processed at ingestion, we only recognize $exception_fingerprint (for custom grouping) or $exception_level

hpouillot (approved):
Nice !

basics/expo/app/profile.tsx
cat-ph: ++ I was about to comment this, I think we should get rid of those `$exception_source` - it's "legacy", we have a plural version now that gets computed from the frames, i.e.: 

https://github.com/PostHog/posthog/blob/e9e1014e0952afdd17eb7211acc90b90eefc8061/rust/cymbal/src/fingerprinting/grouping_rules.rs#L158-L161

cat-ph (approved):

basics/expo/app/profile.tsx
gewenyu99: I'll remove this for now. I don't think this should be as a part of the basic integration skill anyway.

PR #156, opened 2026-05-28
fix: AI assistant loses deployment region context during setup workflow

cristhianrivera:
## What happened

While onboarding PostHog and configuring a self-managed analytics setup, I turned to the PostHog chat agent for help with the manual instrumentation steps. That's when I noticed the agent had no awareness of which cloud region I was on. It consistently defaulted to US Cloud during my setup flow, referencing `us.i.posthog.com` endpoints even though my project lives on EU Cloud. The setup instructions it generated pointed at the wrong ingest host, which caused events to fail silently until I caught the discrepancy manually.

## Root cause

The skill descriptions for all six `instrument-*` omnibus skills, and the base `integration` skill, listed both US and EU host options — but gave the AI no instruction on how to determine which one applied. Without resolution logic, the agent had no way to surface the ambiguity.

## Fix

- All six omnibus skill descriptions (`instrument-integration`, `instrument-product-analytics`, `instrument-error-tracking`, `instrument-feature-flags`, `instrument-llm-analytics`, `instrument-logs`) now instruct the agent to check the `projects-get` MCP response for a `region` field first. If that's unavailable, it must ask the user explicitly: *"Are you on PostHog US Cloud or EU Cloud?"* — and is explicitly forbidden from assuming US Cloud.
- `integration/description.md` gains a new **Region and host URL** section; it previously had no host guidance at all.

## Testing

All 60 existing tests pass (`npm test`).

### DIFF
--- a/transformation-config/skills/integration/description.md
+++ b/transformation-config/skills/integration/description.md
@@ -14,9 +14,13 @@ Follow these steps in order to complete the integration:
 
 The example project shows the target implementation pattern. Consult the documentation for API details.
 
+## Region and host URL
+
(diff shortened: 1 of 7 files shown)


cristhianrivera:
Closing in favor of #217 — same fix, rebased onto the transformation-config/ → context/ rename (which is what made this one conflict) and extended to cover the docs-only integration templates. Thanks!

PR #159, opened 2026-05-29
New Skill: Changelog watch

joethreepwood:
A medium-ambitious skill for the Wizard, based on one of @SaraMiteva [scheduled tasks](https://github.com/PostHog/posthog.com/pull/17101/changes). It's my first time writing a skill, so feedback is appreciated for both me and my compatriot, Claude.

Idea here is that this lets me build something with:

- Regular output (not just a one-off)
- Interactive elements (it asks for competitors)
- Generically reusable (isn't just a PostHog thing)
- Simple methodology (it's only creating text)

My understanding from a demo with @edwinyjlim was that by adding this here, this becomes available to the Wizard. In addition to general feedback, I'd also love to know if there are any next steps to actually get something like this into production and how you'd recommend I could run and make this visible to users.

Once I understand that, I have a bunch of other ideas queued up.

### DIFF
--- a/transformation-config/skills/changelog-watch/description.md
+++ b/transformation-config/skills/changelog-watch/description.md
@@ -1,0 +1,147 @@
+# Changelog watch
+
+Watch a set of named competitors for recent meaningful moves and produce a short internal digest. Generic — works for any company; the user supplies their own positioning and competitor list on first run, and the skill persists the config across runs.
+
+This skill is **read-only against the user's project** — it never edits source files. The only files it writes are the config (on first run) and the optional digest output file.
+
+## Tools used
+
+- **Read / Write** — for `config.json` persistence and (optionally) writing the digest to a markdown file
+- **Bash** — to create the config directory (`mkdir -p ~/.posthog/changelog-watch`)
+- **WebFetch** — to pull competitor changelog and blog pages
+- **WebSearch** — fallback when WebFetch returns a client-rendered shell
+
+## Config file location
+
+Config is stored at **`~/.posthog/changelog-watch/config.json`** (user-level, shared across repos). On every run:
+
+1. Try to read the config.
+2. If it exists, summarise it back to the user and ask whether to run as-is or edit.
+3. If it does not exist, run the first-run interview below.
+
+Emit a `[STATUS]` line on entry so the Wizard's live spinner reflects the current sub-step:
+
+```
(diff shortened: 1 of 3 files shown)


transformation-config/skills/changelog-watch/description.md
```diff
@@ -0,0 +1,147 @@
...
+
+Read it, then summarise to the user in one line:
+
+> Loaded config: **{company}** — {N} categories, {M} competitors. Run as-is, or update first?
+
+If they want to edit, ask which part (positioning / categories / competitors / focus list) and apply only that change. Show the proposed JSON diff to the user before writing.
+
+### If the config is missing — first-run interview
+
+Ask these as free-form chat questions, one at a time. **Do not use `AskUserQuestion`** — this skill ships through PostHog's Wizard and MCP, and Wizard surfaces don't support structured prompts.
```
gewenyu99, line 40:
```suggestion
Ask these as free-form chat questions, one at a time. Use the `mcp__wizard-tools__wizard_ask` tool to surface questions to the users.
```

gewenyu99 (approved):
I think we should ship this and see what happens. I have a few open ended questions:

- Who will be running this, realistically? Is this an internal tool for us or is this something to share with customers?
- I love that this is built for the wizard. Should we have a variant that is for Claude Cowork and ChatGPT? PostHog Code?
- I guess a better question, do you think Claude Cowork with it's ability to automatically rerun these jobs would be a better fit?

gewenyu99 (commented):
As a follow up, I think you'll have sage wisdom and tips related to this we can display while it's running

Image

edwinyjlim:
+1 to @gewenyu99's questions

the skill is well structured and working! ...but it might not be *for* the wizard 

It's my fault. there are some constraints I didn't explain in detail yet @joethreepwood

the wizard has restricted internet access. for security reasons, we have to confine the wizard's online reach in case it downloads something harmful and runs it on the user's laptop. since this skill requires open-ended internet research, it would be more suited to run in Claude Code or Claude Desktop 
 
for your next skill @joethreepwood, I'd pick a more PostHog-specific task so we can deploy it to users via the wizard. anything from [your initial list](https://posthog.slack.com/archives/C09GTQY5RLZ/p1779378375680969) is 🤌

we should still merge and publish this skill to the context mill, because it's great and it'll be available to download and share with the team

PR #163, opened 2026-06-02
fix: pin warlock triage temperature to 0

sarahxsanders:
pin the warlock triage llm calls temp to 0. default temp for haiku is too high, making re-runs and debugging harder because it won't reproduce similar results

temperature 0 makes the verdict reproducible: the same content gets the same answer every run

**can't ship until the warlock PR ships first, we also need to improve the triage model a bit


edwinyjlim (approved):
🥶

PR #169, opened 2026-06-05
update to scan for wrapper functions

edwinyjlim:
updates to scan for wrapper functions around PostHog SDK captures


transformation-config/skills/events-audit/references/4-query.md
```diff
@@ -94,11 +94,21 @@ Set top-level keys on the inventory based on what happened:
 | Query ran but the result was empty (zero matching events in last 30d) | `true` | `"empty result — likely wrong project, but proceeding"` |
 | IN-list was empty (every capture is dynamic) | `true` | `"no resolved event names to query"` |
 | Project couldn't be confirmed in (a) | `false` | `"project mismatch"` |
-| `query-run` errored out (misconfigured project, schema drift, network) | `false` | `"query-run failed: <short reason>"` |
+| `execute-sql` errored out (misconfigured project, schema drift, network) | `false` | `"execute-sql failed: <short reason>"` |
 | No MCP connection at all | `false` | `"MCP unavailable"` |
 
 When `mcp_available: false`, leave every row's `volume_30d: null`, `last_seen: null`, `status: "pending"`. Step 5 reads these flags and renders a disclaimer in place of volume-dependent sections. Step 6 reads `mcp_available` and skips dashboard creation entirely if false.
 
+### g.1. Compute `wrapper_likely`
```
gewenyu99, line 102: :squint: this is an interesting shape. How do we test this? 🤔

transformation-config/skills/events-audit/references/2-scan.md
```diff
@@ -4,20 +4,34 @@ next_step: 3-enrich.md
 
 # Step 2 – Scan capture sites
 
-Find every PostHog capture/identify/group SDK call in the codebase via a single `Grep` and write a base inventory. **Read-only via Grep.** Don't `Read` any source files in this step — file-level enrichment happens in step 3.
+Find every PostHog event-capture call site in the codebase — both **direct SDK calls** (`posthog.capture(...)`) and calls through **typed wrappers** (`captureEvent(...)`, `track(...)`, etc.) — then write a unified base inventory.
```
gewenyu99, line 7: I almost feel like, we can be less descriptive and let Claude trace it themselves. Especially the `**typed wrappers** (`captureEvent(...)`, `track(...)`, etc.)` smells weird to me. 

What we might be better off doing is finding what looks like the wrapper shape and asking the user.

My gut says when the flow looks this descriptive, we'll have many cases where the agent would have found it faster with less guidance.

gewenyu99 (approved):
I'm good with this, but I feel like we should ask the agent to trace PostHog imports to find the wrapper instead of prescribing shapes. It probably will find it faster itself than with suggested shapes that are just guesses.

PR #170, opened 2026-06-05
feat: add pii-bouncer skill

joethreepwood:
## Summary

The context-mill **skill** ("cartridge") for the wizard's PII Bouncer program — companion to **PostHog/wizard#510**. The wizard side is pure engine; this is the packaged-English methodology it loads at runtime.

`pii-bouncer` is a docs-only remediation skill that hardens a frontend against **PII leaking into session replay and autocapture**:

- adds the **`ph-no-capture` class** to sensitive elements (the replay-masking mechanism — distinct from the `data-ph-no-capture` autocapture attribute)
- tightens `session_recording` mask config: adds `maskTextSelector` for sensitive **text** (not masked by default) and confirms `maskAllInputs` isn't disabled (it defaults to `true`, so inputs are already masked — the skill doesn't restate that)
- writes `posthog-pii-bouncer-report.md` (elements masked + why, init changes, reviewed-but-skipped, manual follow-ups)

## Design

- **`type: docs-only`**, single variant `id: all` → skill id `pii-bouncer` (matches the wizard's `skillId`).
- **`shared_docs`** pulls in `session-replay/privacy.md` + `libraries/js/config.md`, bundled as `references/` so the agent uses the authoritative option names and defaults instead of guessing.
- Emits the `[ABORT]` signals the wizard routes verbatim: `no-posthog-js`, `no-init-call`, `no-frontend-templates`.
- Conservative heuristic (type / autocomplete / name·id / label / placeholder / rendered text); bias toward masking. Idempotent — a second run is a no-op.

This is complementary to the new warlock PII rules: warlock stops PII being **sent as event properties**; this skill stops PII being **recorded in the DOM**.

## Test plan

- [x] `pnpm build` produces `dist/skills/pii-bouncer.zip` containing `SKILL.md` + `references/{privacy,config}.md`
- [x] `pii-bouncer` appears in `skill-menu.json`
- [x] `pnpm dev` serves it at `http://localhost:8765`; the wizard (localMcp) resolves + downloads it
- [ ] **Manual e2e (reviewer):** run the wizard's `pii-bouncer` against a workbench app and confirm the masking edits + report

### DIFF
--- a/transformation-config/skills/pii-bouncer/description.md
+++ b/transformation-config/skills/pii-bouncer/description.md
@@ -1,0 +1,141 @@
+# PII Bouncer
+
+This skill hardens a frontend project against **PII leaking into session
+replay and autocapture**. It edits project source: it adds the
+`ph-no-capture` class to sensitive elements and tightens the masking options
+on `posthog.init`. The only file it *creates* is the final report.
+
+This is the prevention layer that complements PostHog's other PII defenses:
+the security scanner stops PII being **sent as event properties**; this skill
+stops PII being **recorded in the DOM** (replay) or **read by autocapture**.
+
+## What posthog-js already does (don't undo it)
+
+Read the bundled reference docs — `references/privacy.md` and
+`references/config.md` — for the authoritative API and defaults. The two facts
+that shape this skill:
+
+- **`maskAllInputs` defaults to `true`.** Every `<input>` value is already
+  masked in replay out of the box. Your job is **not** to "turn on input
+  masking" — it's to (a) confirm nobody disabled it, and (b) cover the
+  surfaces defaults miss.
+- **General text is NOT masked by default.** Sensitive content rendered as
+  text (an SSN on a profile page, an order total, an account balance) is
+  recorded verbatim unless you mask it. This is the main gap to close.
+
+So the value this skill adds is: masking sensitive **text**, blocking
+sensitive **non-input elements**, and verifying the input default is intact —
+not restating defaults.
+
+## The two mechanisms
+
+1. **`ph-no-capture` (CSS class)** — add it to an element and that element is
+   replaced by a same-size placeholder block in replay, **and** autocapture is
+   disabled for it. Use for discrete sensitive elements (a card-number field,
+   a div showing an SSN, a billing summary). This is a class, not an
+   attribute — in JSX use `className="ph-no-capture"`, everywhere else
+   `class="ph-no-capture"`. **Merge** it into any existing class list; never
+   clobber existing classes.
+   - Do **not** confuse this with the `data-ph-no-capture` *attribute*, which
+     is a separate, autocapture-only opt-out. For replay masking, use the
+     **class**.
+
+2. **Init mask config** (`posthog.init(token, { session_recording: { … } })`)
+   — project-wide masking. The high-value option here is `maskTextSelector`
+   (e.g. `"*"` to mask all text, or a scoped selector for sensitive regions).
+   Confirm `maskAllInputs` is not set to `false`. Use the exact option names
+   and defaults from `references/config.md` — do not guess them.
+
+## Workflow
+
+Emit a `[STATUS]` line (see below) as you enter each phase. Before phase 1,
+read `references/privacy.md` and `references/config.md` — they are the source
+of truth for the masking option names and defaults.
+
+1. **Confirm prerequisites.** If `posthog-js` is not a dependency anywhere,
+   emit `[ABORT] no-posthog-js`. Find the `posthog.init(...)` call; if there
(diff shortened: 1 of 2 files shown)


edwinyjlim:
@joethreepwood this is a perfectly constructed context mill skill 👏

i don't even have to read it. you can tell just by the shape and how you're stitching posthog docs as a reference with your skill instructions

transformation-config/skills/pii-bouncer/description.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Workflow
+
+Emit a `[STATUS]` line (see below) as you enter each phase. Before phase 1,
+read `references/privacy.md` and `references/config.md` — they are the source
+of truth for the masking option names and defaults.
+
+1. **Confirm prerequisites.** If `posthog-js` is not a dependency anywhere,
+   emit `[ABORT] no-posthog-js`. Find the `posthog.init(...)` call; if there
+   is none, emit `[ABORT] no-init-call`. Enumerate frontend templates
```
sarahxsanders: for this sentence: `Find the `posthog.init(...)` call; if there
   is none, emit `[ABORT] no-init-call`.`

React projects may not have a literal `posthog.init(...)` call to find, so in this case, it would hit the abort on a valid React project. I'd also have it detect `<PostHogProvider>` and write masking config into its `options={...}` prop instead. Same outcome, framework-correct

claude should be able to fix this easily ^^^

transformation-config/skills/pii-bouncer/description.md
```diff
@@ -0,0 +1,141 @@
...
+
+Conservative bias: **when a signal matches, mask it.** Over-masking only
+over-protects; under-masking leaks PII. Treat an element as sensitive when
+any of these hold:
+
+| Signal | Matches |
+|---|---|
+| `type` | `password`, `email`, `tel` |
+| `autocomplete` | `current-password`, `new-password`, `cc-number`, `cc-csc`, `cc-exp`, `email`, `tel`, `one-time-code`, `ssn` |
+| `name` / `id` | matches `/pass(word)?\|pwd\|card\|cc[-_]?(num\|number)\|cvv\|cvc\|csc\|ssn\|sin\|nin\|pin\|dob\|birth\|tax\|passport\|iban\|account[-_]?(num\|number)/i` |
```
sarahxsanders: the alternatives in this regex have no word boundaries so it matches anywhere inside another word. for example passenger matches pass, spinner matches pin, etc. could get messy with bury sensitive stuff and add to token waste

if you wrap each alternative in `\b...\b` it should help:

`/\b(pass(word)?|pwd|card|cc[-_]?(num|number)|cvv|cvc|csc|ssn|sin|nin|pin|dob|birth|tax|passport|iban|account[-_]?(num|number))\b/i` for example

sarahxsanders (changes requested):
one thing that could potentially be breaking for React projects, then you're good for ✅

transformation-config/skills/pii-bouncer/description.md
joethreepwood: Fixed in fd9c087. Step 1 now looks for either a `posthog.init(...)` call **or** a `<PostHogProvider>`, and only emits `[ABORT] no-init-call` if neither exists. Step 4 writes the masking config into whichever it found — the `session_recording` options on `init`, or the `options={{…}}` prop on the provider. (Also broadened the wizard's abort copy to match: PostHog/wizard#510.)

transformation-config/skills/pii-bouncer/description.md
joethreepwood: Good catch — word-bounded in fd9c087: `/\b(pass(word)?|pwd|card|cc[-_]?(num|number)|cvv|cvc|csc|ssn|sin|nin|pin|dob|birth|tax|passport|iban|account[-_]?(num|number))\b/i`, with a note that it's bounded so `pass` doesn't match `passenger`, `pin` doesn't match `spinner`, etc.

joethreepwood:
Thanks @sarahxsanders — both points addressed in `fd9c087`:

- **React `<PostHogProvider>`** — step 1 now treats either a `posthog.init(...)` call **or** a `<PostHogProvider>` as a valid init site, and only emits `[ABORT] no-init-call` when neither exists (no more false abort on valid React apps). Step 4 writes the masking into whichever it found — `session_recording` on `init`, or the `options={{…}}` prop on the provider. The wizard's abort copy was broadened to match (PostHog/wizard#510).
- **Word-bounded the name/id regex** — `/\b(pass(word)?|pwd|card|…)\b/i`, so `pass` no longer matches `passenger`, `pin` no longer matches `spinner`, etc.

Also added a reciprocal breadcrumb back to the wizard's `abort-cases.ts` so the `[ABORT]` contract is explicit on both sides (`e7c5bfa`).

PR #171, opened 2026-06-05
Swift: recommend hardcoded token over env-only config

daniloc:
## Problem

User feedback from an iOS integration:

> The `PostHogEnv` `fatalError` pattern from SKILL.md crashes production iOS archive builds because Xcode scheme env vars are not available at archive time. Had to add a production fallback (default values).

Root cause: **Xcode scheme environment variables are injected only when launching from Xcode (debug/simulator). They are absent in Archive/Release builds** (TestFlight, App Store). The previous `PostHogEnv` enum read the token from `ProcessInfo.processInfo.environment` and called `fatalError` when missing — so any distribution build crashed on launch.

## Fix

A PostHog **project token is a public client-side key** designed to ship in the app binary. So there's no secret to keep out of the binary, and env-var injection only works in one narrow mode (launch-from-Xcode) anyway. The token is now **set directly in `PostHogConfig`** — no `ProcessInfo` reading, no enum ceremony.

- `transformation-config/commandments.yaml` — replace the `fatalError` Swift commandment with hardcode-the-token guidance + an explicit "do NOT read the token from scheme env vars" rule explaining the archive-build limitation. Regenerates the Swift `SKILL.md` guidelines.
- `basics/swift/.../BurritoConsiderationClientApp.swift` — drop the `PostHogEnv` env-reading enum; use plain `posthogProjectToken` / `posthogHost` constants.
- `basics/swift/README.md` — updated config section and code sample to match.


gewenyu99 (approved):
LOL nice

PR #178, opened 2026-06-08
feat(cli): cli overhaul

sarahxsanders:
CLI OVERHAULLLLLLLLLLLLLL 🤠

two things happening here:

1. adds `cli:` block to skill `config.yaml` to tell the wizard how a
   skill's command should appear
2. removes `audit-3000` in favor of a proper `audit` skill family:
   `audit events`, `audit feature-flags` etc.

## `cli:` block

each skill has the option to declare:

```yaml
cli:
  role: command          # command | skill | internal
  parentCommand: audit   # optional, nests this under a parent
  command: events        # the word the user types
  recommended: true      # optional, the pre-highlighted leaf in a family
```

the build parses + validates it (kebab-case, length limits, reserved-word
collisions, etc.) and emits the entries into skill-menu.json under
cliEntries. the wizard fetches skill-menu.json at runtime and uses
that array to register its command surface – no baked snapshot, no
build-time coupling between the two repos.

cli-manifest.json still ships as a standalone file for one release as
back-compat with the still-baked older wizard. it gets deleted in a
follow-up once that release ages out.

full schema and semantics in CONTRIBUTING.md.

## how the commands come together 

1. flat command

```yaml
cli:
  role: command
  command: revenue-analytics
```

command: `wizard revenue-analytics`

2. command nested under a family

```yaml
cli:
  role: command
  parentCommand: audit
  command: events
```

command: `wizard audit events`

3. recommended leaf in a family

```yaml
cli:
  role: command
  parentCommand: audit
  command: all
  recommended: true
```

command: `wizard audit` then Enter runs this one. the picker still opens
(so people can see the other audit options + opt in), recommended just
sorts this to the top so the common case is one keystroke. only one leaf
per family gets to be the recommended one, and the build will yell at
you if two try to claim it.

4. skill

```yaml
cli:
  role: skill
```

command: `wizard skill <skill-id>`

5. internal

```yaml
cli:
  role: internal
```
call with `--skill=<id>` for dev

## flat vs family

a command is flat when there's only one option today, and a family when
the user has to pick between a few.

the rule of thumb: don't build the family shape before you need it. no
`wizard migrate <vendor>` while there's only one vendor – that's just
abstraction for its own sake. when a second option shows up, that's the
moment to restructure into a family. the migration steps for that
restructure (and the breaking-change call-out) are in
`CONTRIBUTING.md`.

## sister PR

https://github.com/PostHog/wizard/pull/617

something to note: when the wizard upgrades to a release with these
changes, we need to run `pnpm docs:cli` to refresh the committed
`cli.md` (I can do this).

release order (IMPORTANT):
1. merge this PR first
2. cut a context mill release
3. merge wizard PR and cut a release

CANT SHIP WIZARD PR FIRST — the wizard fetches `skill-menu.json` at
runtime, so the new shape (with `cliEntries`) has to exist in a
published release before the wizard release goes out.

## testing
- `npm test` all passing
- `npm run build` green; `dist/skills/skill-menu.json` has `cliEntries`
with the expected entries and recommended flags; `dist/skills/cli-manifest.json`
still emits the same shape for back-compatibility
- build context mill locally, build wizard to verify it pulls commands at runtime, test different skills:
CleanShot 2026-06-17 at 10 19 58


CLAUDE.md
gewenyu99, line 1: Move this to agents.md so all agents will read i think

context/cli-manifest.schema.json
gewenyu99, line 1: what iz this file?

package.json
```diff
@@ -16,6 +16,7 @@
     "security-scan:skills": "node scripts/scan-warlock.js dist/skills"
   },
   "devDependencies": {
+    "ajv": "^8.20.0",
```
gewenyu99: Why this package specifically? :thinking_face:

gewenyu99 (commented):
an initial fly by, will return to this after looking at wizard PR, too.  
  
Minorly concerned about tight coupling wizard builds to context mill for command registering.  
  
My initial thoughts is that commands are not context and are too wizard specific. What part of the command is strictly context/information? And what parts are just wizard specific?

context/cli-manifest.schema.json
sarahxsanders: a rulebook for `cli-manifest.json` to spell out the shape every entry has to follow:
- fields
- their types
- allowed role values
- naming rules
etc.

published next to manifest so the wizard can check the manifest it pulls. this catches drift at build instead of at production

It's also what the ajv check in the following comment validates against

package.json
sarahxsanders: using it to validate `cli-manifest.json` against `cli-manifest.schema.json` at build time. this means it fails build here and doesn't impact wizard

picked this package because the schema is JSON and `ajv` is what claude recommended for that

also see above

context/cli-manifest.schema.json
sarahxsanders: this could also be way too complex so feel free to tell me I am overthinking it/there's a cleaner way to validate

package.json
sarahxsanders: ignore we ripping out

scripts/lib/skill-generator.js
```diff
@@ -7,6 +7,102 @@
  * - Commandments (based on tags)
  */
 
+/**
+ * Optional `cli:` block in a skill's `config.yaml`.
```
gewenyu99: I would honestly throw this in a markdown. This comment block is just a little tooo chonky

scripts/lib/skill-generator.js
```diff
@@ -7,6 +7,102 @@
...
+ *     role: command
+ *     parentCommand: audit
+ *     command: feature-flags
+ *
+ *   # 3. Recommended leaf inside a family (`wizard audit` runs this on Enter)
+ *   cli:                                          →  wizard audit all
+ *     role: command                                  Pre-highlighted in the
+ *     parentCommand: audit                           family picker so
+ *     command: all                                   `wizard audit` → Enter
+ *     recommended: true                              runs this leaf.
```
gewenyu99: Should this be called `default`

scripts/lib/skill-generator.js
```diff
@@ -21,6 +117,147 @@ function loadYaml(configPath) {
     return yaml.load(content);
 }
 
+const CLI_ROLES = ['command', 'skill', 'internal'];
+
+// Naming convention enforcement — see context-mill/CONTRIBUTING.md and the
```
gewenyu99: Maybe move this to its own file. Great that we're enforcing, not super relevant to the flow of this which makes it just a lil harder to read

gewenyu99:
https://gist.github.com/gewenyu99/e60f12296774548c1e13364e33576962

Diffs

gewenyu99 (approved):
LGTM mate

PR #194, opened 2026-06-22
feat: agentic detection

ablaszkiewicz:
For some reason wizard started hallucinating when I hooked up agentic detection. This fixes it

Also it explains how to use combined `process` instead of `inject` and `upload` separately


hpouillot (approved):

cat-ph (approved):

gewenyu99 (approved):
Cool with me. Future tip:

Value labels like `#### Tips` are likely not helpful for agents. Especially sonnet. Probably use a more concrete label that carries advice. Remove these filler, wishy-washy terms might make your skills better.

ablaszkiewicz:
@gewenyu99 - thanks for that. Can we do it as a part of future changes? I don't want to re-test everything again after touching context-mill. I know I owe you CICD tests - I still have that on my list 🙏

PR #157, opened 2026-05-28
feat: add data-warehouse-source-setup skill

Gilbert09:
## Summary

Adds the `data-warehouse-source-setup` skill that powers the PostHog wizard's new `warehouse` program (companion PR: [PostHog/wizard#488](https://github.com/PostHog/wizard/pull/488)).

The wizard detects which data sources a project uses and passes them (kind + creation mode) into the agent prompt. This skill is the agent playbook for connecting them:

- **`in-cli` sources** (databases, API-key SaaS): discover required fields via the PostHog MCP `external-data-sources-wizard`, collect credentials in a single batched `wizard_ask` (never reads `.env` values), validate/list tables via `external-data-sources-db-schema`, then create with `external-data-sources-create`.
- **`deep-link` sources** (OAuth): emit the pre-filled `new-source?kind=` URL for the user to finish in the app.
- **CI / non-interactive**: falls back to deep-link for everything.
- `[ABORT]` reasons (`No data source detected`, `Source creation failed`) match the wizard's abort-case handling.

`docs-only` skill; bundles the Postgres/MySQL/Snowflake/BigQuery/Stripe + sources-overview reference docs. Generates to skill id `data-warehouse-source-setup`.

## Test plan

- [x] `pnpm test:skills` — 38/38 pass
- [x] Skill generates cleanly (id resolves to `data-warehouse-source-setup`, all template placeholders resolved, reference docs fetched)
- [x] `scan-prompt-injection.sh` — PASSED

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #206, opened 2026-06-30
fix(data-warehouse): reduce source-setup wizard friction

Gilbert09:
## Problem

The wizard's `warehouse` program (`npx @posthog/wizard`) drives an AI agent through the `data-warehouse-source-setup` skill added in #157. In production a large share of runs were aborting. Reading the agent's own post-run remarks surfaced that the skill guidance itself was a major contributor — so this updates the skill.

## Changes

- **Always pass `source_type` to `external-data-sources-wizard`.** The skill told the agent to call it but not to filter; the unfiltered response is hundreds of KB and was blowing the agent's context budget on nearly every run (the single biggest cause). Companion API change adds the `source_type` param: PostHog/posthog#67018.
- **Fix the `wizard_ask` batching contradiction.** Tenet 2 said both "single call" and "one source, one prompt" with no cancellation guidance, which the agent repeatedly got stuck on. Rewritten: batch a source's fields into one call (≤8), one call per source is fine, sequential follow-ups allowed for answer-dependent questions, and a cancelled/timed-out ask is free → fall back to deep-link.
- **Add a secret-reference tenet.** Sensitive `wizard_ask` answers come back as `{ secretRef }`, which the PostHog data-warehouse tools reject. The skill now says to collect credentials destined for those tools as plain `text` and reserve `sensitive` for values only written to `.env`.
- **Add a pre-flight gotchas section** — network reachability (localhost/private hosts rejected, egress allowlisting), Supabase session-pooler specifics (host/username/port 6543/db password), and per-source key types (Stripe restricted key, Sentry internal-integration token, etc.) so credentials validate on the first attempt.

Follow-up to #157.

## Test plan

I'm an agent (Claude Code); changes are docs-only to the skill template.

- `pnpm test:skills` — 70/70 pass
- `pnpm build` (node 22) — `data-warehouse-source-setup` generates cleanly (`✓`), all placeholders resolve

The companion `wizard_ask` cap fix lives in PostHog/wizard (a cancelled ask no longer counts against the per-run cap), and the API `source_type` filter + secretRef rejection in PostHog/posthog#67018.

### DIFF
--- a/context/skills/data-warehouse-source/description.md
+++ b/context/skills/data-warehouse-source/description.md
@@ -17,7 +17,7 @@ Consult the PostHog data warehouse source docs above for source-specific field r
 
 You have the PostHog MCP server and the wizard's local tools available:
 
-- **`mcp__posthog-wizard__external-data-sources-wizard`** — returns the required fields per source type. **Always call this for a source kind before creating it** — never guess field names.
+- **`mcp__posthog-wizard__external-data-sources-wizard`** — returns the required fields per source type. **Always call this for a source kind before creating it** — never guess field names. **Pass `source_type` with the kind(s) you need** (e.g. `source_type: "Postgres"`, or comma-separated `"Postgres,Stripe"`). The unfiltered response describes every source and is hundreds of KB — large enough to blow your context budget — so never call it without `source_type`.
 - **`mcp__posthog-wizard__external-data-sources-db-schema`** — validates credentials and lists the tables available for sync. Use this for database sources before creating.
 - **`mcp__posthog-wizard__external-data-sources-create`** — creates the source. Follow its input schema exactly for the `payload` and `schemas` shape; the tool definition is the source of truth.
 - **`mcp__wizard-tools__check_env_keys`** — tells you which `.env` keys EXIST. It never returns values.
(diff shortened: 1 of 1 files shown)


context/skills/data-warehouse-source/description.md
```diff
@@ -27,13 +27,23 @@ You have the PostHog MCP server and the wizard's local tools available:
...
-5. **Don't modify project code.** This skill connects external data; it does not edit the user's application. Make no source-code changes.
+5. **Respect the mode.** Only collect credentials and create `in-cli` sources. For `deep-link` sources, provide the URL and stop — do not try to collect OAuth tokens.
+
+6. **Don't modify project code.** This skill connects external data; it does not edit the user's application. Make no source-code changes.
+
+## Pre-flight: credential gotchas that cause most failures
+
+Surface these **before** collecting credentials — they're the top reasons setup fails on the first try, and a failed attempt wastes a `wizard_ask`.
+
+- **The host must be reachable from PostHog's network.** `localhost`, `127.0.0.1`, and private/RFC-1918 hosts (`10.x`, `192.168.x`, `172.16–31.x`) are rejected — PostHog connects from its own infrastructure, not the user's machine. Serverless/managed Postgres (Neon, Supabase, RDS behind strict rules) often also needs PostHog's egress IPs allowlisted first. If the database isn't publicly reachable, go straight to the deep-link path instead of collecting credentials that can't validate.
```
gewenyu99, line 44: What of these prompts can you write an MCP tool and/or script for the agent to call once and check deterministically? Just a useful thought

gewenyu99 (approved):

PR #208, opened 2026-06-30
Use "project API token" terminology over "API key"

gewenyu99:
Aligns the wording the wizard emits with PostHog's **project API token** terminology. The project token (the public `phc_…` ingestion key) was being called an "API key" in several skill descriptions and one example app — inconsistent with the project-token convention used everywhere else, and the source of the wizard emitting `POSTHOG_API_KEY` instead of a token.

## What changed
- **Omnibus `instrument-*` skills** (all six): the credential step now asks for the **project API token** rather than the "API key".
- **`mcp-analytics`**: env var `POSTHOG_PROJECT_API_KEY` → `POSTHOG_PROJECT_TOKEN`; prose → "project API token".
- **PHP example app**: `POSTHOG_API_KEY` → `POSTHOG_PROJECT_TOKEN` (the lone example app that used the old name).
- **`commandments.yaml`** (Rails): "Store API key…" → "Store the project API token…".
- Renamed incidental local variables that hold a project token to `projectToken` (js-web, expo, react-native) for consistency with the node example.

## Deliberately left untouched
- Genuine **personal API key** references (source-map upload / error tracking, local-evaluation keys, query-API personal keys) — "API key" is correct there.
- The SDKs' own `apiKey` / `api_key` / `PostHogConfig(apiKey:)` **parameter names**, which are fixed by the SDK signatures — renaming them would break the code.

## Why
Feedback that the wizard's generated setup labels the project token as an "API key" instead of a token.

`npm test` (92 passing) and `npm run build` both green.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C043VJ93L3B/p1782789346955049?thread_ts=1782789346.955049&cid=C043VJ93L3B)*

### DIFF
--- a/context/commandments.yaml
+++ b/context/commandments.yaml
@@ -170,7 +170,7 @@ commandments:
     - "capture_exception takes POSITIONAL args: PostHog.capture_exception(exception, distinct_id, additional_properties) — do NOT use keyword args"
     - "Define posthog_distinct_id on the User model for automatic user association in error reports — posthog-rails auto-detects by trying: posthog_distinct_id, distinct_id, id, pk, uuid (in order)"
     - "For ActiveJob user association, use the class-level DSL `posthog_distinct_id ->(user) { user.email }` or pass user_id: in a hash argument"
-    - Store API key in Rails credentials or environment variables, never hardcode
+    - Store the project API token in Rails credentials or environment variables, never hardcode
     - "For frontend tracking alongside posthog-rails, add the posthog-js snippet to the layout template — posthog-js handles pageviews, session replay, and client-side errors while posthog-ruby handles backend events, server errors, feature flags, and background jobs"
 
   hogql:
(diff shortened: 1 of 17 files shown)


edwinyjlim:
@gewenyu99 I think you need to one more agent sweep. It should be `PostHog project token` not `PostHog project API token` (drop the API)

gewenyu99:
Oh no API either?

gewenyu99:
Done in 26ea636 — dropped the "API" so it now reads **"project token"** (not "project API token") across the omnibus `instrument-*` skills, `mcp-analytics`, and the Rails commandment. The MCP `api_token` field name in code and the "personal API key" references are left as-is.

lucasheriques (approved):
Looks good, and the scoping (leaving personal API key references and SDK apiKey parameters alone) is right. One follow-up: the posthog.com MCP analytics docs still use POSTHOG_PROJECT_API_KEY in their snippets, so after this merges the skill and the published docs will disagree on the env var name. I'll raise a companion docs change to align them.

PR #211, opened 2026-07-01
feat: create dashboards and insights before code in wizard integration flow

rafaeelaudibert:
## What

Reorders the phases of the `integration` skill (the wizard's default flow, program id `posthog-integration`) so the PostHog **dashboard and insights are created from the event plan _before_ any tracking code is written**, instead of last. The result: the analytics we want to answer drive the instrumentation, rather than trailing it.

## Why

Today the flow is `begin (plan) → edit (code) → revise → conclude (create dashboard + report)`. The full "what to instrument" decision already lives in `.posthog-events.json` after phase 1, so there's a natural point to branch: build the dashboard from the plan first, then implement against it.

## Phase changes

New chain: `1-begin → 2-dashboard → 3-edit → 4-revise → 5-conclude`.

| Phase | Change |
|---|---|
| `1-begin` | `next_step` → `2-dashboard`; intro sequence reworded |
| `2-dashboard` *(new)* | Creates `Analytics basics (wizard)` + up to 5 insights from `.posthog-events.json`, using the exact planned event names. States explicitly that the events have no data yet (expected). Records the dashboard/insight ids for the final phase; does **not** emit `[DASHBOARD_URL]` |
| `3-edit` (was `2-edit`) | No-rename guard: implement each event with the exact planned `event_name` so code and the pre-built insights line up; if a change is unavoidable, update `.posthog-events.json` and flag it |
| `4-revise` (was `3-revise`) | Plan-adherence check: every planned event is implemented, with the exact name, in the file the plan named; deviations reconciled in code **and** on the dashboard via MCP |
| `5-conclude` (was `4-conclude`) | No longer *creates* the dashboard — **locates** it via the PostHog MCP (by noted id, falling back to name / `(wizard)` tag), emits `[DASHBOARD_URL]`, writes the report, deletes `.posthog-events.json`. Also fixes a stale `description` frontmatter |

## Wizard impact

**None required.** `[DASHBOARD_URL]` is still emitted from the final phase, so the wizard's success-summary parsing is unchanged. The `[STATUS]`/`[ABORT]`/`[DASHBOARD_URL]` markers aren't parsed in this repo at all — only wizard-side.

## Scope

- **In scope:** `integration` skill only.
- **Out of scope:** `events-audit` shares the "dashboard-last" pattern but audits _existing_ instrumentation (no implementation phase follows), so dashboard-last is correct there. `self-driving` doesn't create dashboards. Neither changes.

## Known tradeoff

Creating the dashboard before code means an aborted run can leave an orphan `(wizard)`-tagged dashboard whose events were never instrumented. The tag makes such orphans trivial to find and delete; abort is wizard-side and can't be cleaned up from skill content.

## Verification

- `npm test` — 92/92 pass.
- `npm run build` — succeeds (114 skills); generated `SKILL.md` workflow list shows the 5 phases in the new order (`1-begin` marked "Start here") and every `Upon completion, continue with:` link chains correctly.
- End-to-end wizard run against a sample app is the one check not doable from this repo — worth a spot-check that the dashboard/insights appear before file edits and `[DASHBOARD_URL]` resolves.


rafaeelaudibert:
Closing because the new wizard architecture will solve this in a smarter way

PR #212, opened 2026-07-03
fix: deliver javascript_web/node commandments to Next.js skills (tag mismatch) + align examples

gewenyu99:
## What changed

Stared at CI really hard again. There's some guidance for JS web and JS node not inhereted by frameworks. This makes them sometimes capture PII etc.

## Delivered guidance per runtime

The `## Framework guidelines` section of each generated `SKILL.md`.

### Next.js App Router — full-stack (`javascript_web` + `javascript_node`)

```
- For Next.js 15.3+, initialize PostHog in instrumentation-client.ts for the simplest setup
- For feature flags, use useFeatureFlagEnabled() or useFeatureFlagPayload() hooks - they handle loading states and external sync automatically
- Add analytics capture in event handlers where user actions occur, NOT in useEffect reacting to state changes
- Do NOT use useEffect for data transformation - calculate derived values during render instead
- Do NOT use useEffect to respond to user events - put that logic in the event handler itself
- Do NOT use useEffect to chain state updates - calculate all related updates together in the event handler
- Do NOT use useEffect to notify parent components - call the parent callback alongside setState in the event handler
- To reset component state when a prop changes, pass the prop as the component's key instead of using useEffect
- useEffect is ONLY for synchronizing with external systems (non-React widgets, browser APIs, network subscriptions)
- Remember that source code is available in the node_modules directory
- Check package.json for type checking or build scripts to validate changes
- When a reverse proxy is configured, both /static/* AND /array/* must route to the assets origin (us-assets.i.posthog.com or eu-assets.i.posthog.com).
- posthog-js is the JavaScript SDK package name
- posthog.init() MUST be called before any other PostHog methods (capture, identify, etc.)
- posthog-js is browser-only — do NOT import it in Node.js or server-side contexts (use posthog-node instead)
- Autocapture is ON by default with posthog-js (tracks clicks, form submissions, pageviews). Do NOT disable autocapture unless the user explicitly requests it.
- NEVER send PII in posthog.capture() event properties — no emails, full names, phone numbers, physical addresses, IP addresses, or user-generated content
- PII belongs in posthog.identify() person properties (email, name, role), NOT in capture() event properties
- Call posthog.identify(userId, { email, name, role }) on login AND on page refresh if the user is already logged in
- Call posthog.reset() on logout to unlink future events from the current user
- For SPAs without a framework router, capture pageviews with posthog.capture($pageview) or use the capture_pageview history_change option in init for History API routing
- posthog-node is the Node.js server-side SDK package name – do NOT use posthog-js on the server
- Include enableExceptionAutocapture: true in the PostHog constructor options
- Add posthog.capture() calls in route handlers for meaningful user actions – every route that creates, updates, or deletes data should track an event with contextual properties
- Add posthog.captureException(err, distinctId) in the application's error handler (e.g., Express error middleware, Fastify setErrorHandler, Koa app.on('error'))
- In long-running servers, the SDK batches events automatically – do NOT set flushAt or flushInterval unless you have a specific reason to
- For short-lived processes (scripts, CLIs, serverless), set flushAt to 1 and flushInterval to 0 to send events immediately
- Reverse proxy is NOT needed for server-side Node.js – only client-side JavaScript needs a proxy to avoid ad blockers
```

### Astro (static) — web only (`javascript_web`)

```
- Always use the is:inline directive on PostHog script tags to prevent Astro from processing them and causing TypeScript errors
- Use PUBLIC_ prefix for client-side environment variables in Astro (e.g., PUBLIC_POSTHOG_PROJECT_TOKEN)
- Create a posthog.astro component in src/components/ for reusable initialization across pages
- Import the PostHog component in a Layout and wrap all pages with that layout
- Remember that source code is available in the node_modules directory
- Check package.json for type checking or build scripts to validate changes
- When a reverse proxy is configured, both /static/* AND /array/* must route to the assets origin (us-assets.i.posthog.com or eu-assets.i.posthog.com).
- posthog-js is the JavaScript SDK package name
- posthog.init() MUST be called before any other PostHog methods (capture, identify, etc.)
- posthog-js is browser-only — do NOT import it in Node.js or server-side contexts (use posthog-node instead)
- Autocapture is ON by default with posthog-js (tracks clicks, form submissions, pageviews). Do NOT disable autocapture unless the user explicitly requests it.
- NEVER send PII in posthog.capture() event properties — no emails, full names, phone numbers, physical addresses, IP addresses, or user-generated content
- PII belongs in posthog.identify() person properties (email, name, role), NOT in capture() event properties
- Call posthog.identify(userId, { email, name, role }) on login AND on page refresh if the user is already logged in
- Call posthog.reset() on logout to unlink future events from the current user
- For SPAs without a framework router, capture pageviews with posthog.capture($pageview) or use the capture_pageview history_change option in init for History API routing
```

### React Native — mobile (neither)

```
- posthog-react-native is the React Native SDK package name
- Use react-native-config to load POSTHOG_PROJECT_TOKEN and POSTHOG_HOST from .env (variables are embedded at build time, not runtime)
- react-native-svg is a required peer dependency of posthog-react-native (used by the surveys feature) and must be installed alongside it
- Place PostHogProvider INSIDE NavigationContainer for React Navigation v7 compatibility
- Remember that source code is available in the node_modules directory
- Check package.json for type checking or build scripts to validate changes
```

### DIFF
--- a/context/commandments.yaml
+++ b/context/commandments.yaml
@@ -2,12 +2,18 @@
 # Tag-based rules that get folded into skill descriptions
 # Skills collect all commandments matching their tags
 
-# Runtime implications: a skill tagged with a key below also collects the
-# implied tags' commandments. Keeps per-runtime rules (javascript_web /
-# javascript_node) in ONE place instead of repeating them on every
-# framework's tag list. One level deep — implications do not chain.
-tag_implications:
-  react: [javascript_web]        # react-native does NOT carry `react`, so mobile stays out
+# Tag taxonomy — child tags additively inherit their parents' commandments.
+# `javascript_web` / `javascript_node` are runtime categories ("a generic JS
+# web/server framework"); framework tags declare membership in them, and the
+# categories themselves roll up to the `javascript` root. Resolution is
+# transitive (react -> javascript_web -> javascript) and EXCLUSIVELY additive:
+# expansion only ever adds rule sets, never removes or overrides anything.
+# A variant with an unusual runtime can also declare membership directly by
+# listing the category tag in its own tags (see react-router-7-framework).
+tag_parents:
+  javascript_web: [javascript]
+  javascript_node: [javascript]
+  react: [javascript_web]
   vue: [javascript_web]
   angular: [javascript_web]
   astro: [javascript_web]
@@ -17,7 +23,6 @@ tag_implications:
   astro-ssr: [javascript_node]
   astro-hybrid: [javascript_node]
   tanstack-start: [javascript_node]
-  ssr: [javascript_node]         # JS server-rendered frameworks (only JS skills carry `ssr`)
 
 commandments:
   javascript:
--- a/context/commandments.yaml
+++ b/context/commandments.yaml
@@ -45,6 +45,7 @@ commandments:
     - Autocapture is ON by default with posthog-js (tracks clicks, form submissions, pageviews). Do NOT disable autocapture unless the user explicitly requests it.
     - NEVER send PII in posthog.capture() event properties — no emails, full names, phone numbers, physical addresses, IP addresses, or user-generated content
     - PII belongs in posthog.identify() person properties (email, name, role), NOT in capture() event properties
+    - Every catch block in files you instrument must report the error — call posthog.captureException(err) inside the catch, in addition to (not instead of) the existing handling
     - Call posthog.identify(userId, { email, name, role }) on login AND on page refresh if the user is already logged in
     - Call posthog.reset() on logout to unlink future events from the current user
     - For SPAs without a framework router, capture pageviews with posthog.capture($pageview) or use the capture_pageview history_change option in init for History API routing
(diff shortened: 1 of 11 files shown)


context/commandments.yaml
```diff
@@ -1,10 +1,14 @@
...
 # Tag-based rules that get folded into skill descriptions
-# Skills collect all commandments matching their tags
+# Skills collect all commandments matching their tags. Runtime categories
+# (javascript_web / javascript_node) are ordinary tags; each skill declares
+# the ones it belongs to in its own config (see context/skills/*/config.yaml).
 
 commandments:
+  # Generic JavaScript rules — every JS skill carries the `javascript` tag.
   javascript:
-    - When a reverse proxy is configured, both /static/* AND /array/* must route to the assets origin (us-assets.i.posthog.com or eu-assets.i.posthog.com).
```
gewenyu99, line 7: This is currently not useful or usable. Sometimes the model tries very hard to figure this out.

example-apps/next-app-router/src/app/api/auth/login/route.ts
```diff
@@ -24,7 +24,6 @@ export async function POST(request: Request) {
     distinctId: username,
     event: 'server_login',
     properties: {
-      username: username,
```
gewenyu99, line 27: This will upset our own rubric lol

edwinyjlim (approved):

PR #213, opened 2026-07-06
feat(swift): XcodeGen support for the Swift integration

sarahxsanders:
Pairs with PostHog/wizard#803 (XcodeGen detection) — should land before or with that wizard release.

In XcodeGen projects `project.pbxproj` is a build artifact; the next `xcodegen generate` wipes direct edits. This adds a commandment steering the agent to `project.yml` instead, an `example-apps/swift-xcodegen` reference app, and per-example `allow` skip-pattern overrides so the example's `project.yml` survives the global `.yml` skip.

Out of scope: Tuist (same hazard, defer until one shows up).


example-apps/swift-xcodegen/BurritoStand/BurritoStandApp.swift
```diff
@@ -0,0 +1,27 @@
+import SwiftUI
+import PostHog
+
+// PostHog configuration.
+//
+// The project token is a PUBLIC client-side key — it is designed to ship in the
+// app binary, so hardcoding it here is safe and is the recommended approach for
+// iOS. Replace the placeholder below with your project token from
+// https://app.posthog.com/project/settings.
+private let posthogProjectToken = "<your-project-token>"
```
gewenyu99, line 10: This is sus. Prob don't hardcode either of these in the example? not sure

gewenyu99 (approved):

gewenyu99 (commented):
Oh, I think there's another glob you missed in utils.ts:30

example-apps/swift-xcodegen/BurritoStand/BurritoStandApp.swift
sarahxsanders: the other swift app does this too which is where this pattern came from, so I'll fix in both

sarahxsanders: okay wait, not sure if I SHOULD change this, as the Swift app says it's recommended? https://github.com/PostHog/context-mill/blob/e5fd20554ac8598c2bc75a542f28c77fded88be5/example-apps/swift/BurritoConsiderationClient/BurritoConsiderationClientApp.swift#L13

maybe @daniloc can weigh in as I think? iOS was initially set up by him

daniloc: yeah, if we don't do it this way, it's not properly folded into production builds. this is the way

daniloc (approved):
ship it and see if anyone gripes, seems like the right start to me

PR #217, opened 2026-07-09
fix: require AI to resolve PostHog region instead of defaulting to US Cloud

cristhianrivera:
## What happened

While configuring a self-managed PostHog setup, I noticed the chat agent had no
awareness of which cloud region I was on. It consistently defaulted to US Cloud,
referencing `us.i.posthog.com` even though my project lives on EU Cloud — so the
generated setup pointed at the wrong ingest host and events failed silently until
I caught it manually.

## Root cause

The `instrument-*` omnibus skills and the base `integration` skill listed both US
and EU host URLs but gave the agent no logic for choosing between them, so it had
no way to surface the ambiguity.

## Fix

- All six omnibus skill descriptions now instruct the agent to read the `region`
  field from the `projects-get` MCP response first, and — if unavailable — to ask
  the user explicitly: *"Are you on PostHog US Cloud or EU Cloud?"* It is told not
  to assume US Cloud.
- The `integration` skill templates gain a **Region and host URL** section. This
  covers the base template **and** the docs-only templates (generic, Elixir, Go),
  so Flutter/Elixir/Go integration skills get the same guidance.

## Testing

`pnpm test` → 103 passing. `pnpm build` clean; verified the region guidance lands
in all 42 built integration + omnibus skills.

## Note

This supersedes #156, which was opened before the `transformation-config/ → context/`
rename and no longer merges cleanly. Rebased onto current `main` and extended to
cover the docs-only templates that #156 missed.

### DIFF
--- a/context/skills/integration/description-docs-only.md
+++ b/context/skills/integration/description-docs-only.md
@@ -15,10 +15,6 @@ This skill helps you add PostHog analytics to {display_name} applications using 
 
 {references}
 
-## Region and host URL
-
-PostHog has two cloud regions. Check the `projects-get` MCP response for a `region` field — `US` maps to `https://us.i.posthog.com`, `EU` maps to `https://eu.i.posthog.com`. If the region is not available from the MCP response or from existing project configuration (e.g. an already-set `POSTHOG_HOST` env var), ask the user: "Are you on PostHog US Cloud or EU Cloud?" Do not assume US Cloud.
-
 ## Key principles
 
 - **Environment/configuration values**: Always use environment variables or platform configuration files for PostHog keys. Never hardcode them.
--- a/context/skills/integration/description-docs-only.md
+++ b/context/skills/integration/description-docs-only.md
@@ -15,6 +15,10 @@ This skill helps you add PostHog analytics to {display_name} applications using 
 
 {references}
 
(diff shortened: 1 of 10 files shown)


cristhianrivera:
@gewenyu99 — you've been the main hand on these skill descriptions lately, so tagging you for a look. This is a rebase of #156 onto the context/ rename, plus the docs-only templates it missed. Small, self-contained (skill descriptions only) and CI-green. Happy to adjust wording or scope if you'd prefer it narrower.

context/skills/integration/description-docs-only.md
gewenyu99, line 1: Can you revert changes to the `skills/integration` folder? They're passed region through a different path.

gewenyu99 (changes requested):
Hey, thanks for making these changes! 

Gonna ask you to revert this for the `skills/integration` files, they're passed project through another path when used. The rest look great (and are the ones that end up in PostHog Code)

cristhianrivera:
Done. Reverted the region section from all the skills/integration templates (base + the docs-only ones), kept it in the six omnibus instrument-* descriptions. Pushed as c9dc3c6.

gewenyu99 (approved):

PR #218, opened 2026-07-09
feat: reduce warlock false positives

sarahxsanders:
this should cut down on some of the more obvious and annoying false positives with the warlock


gewenyu99 (approved):
LOL

PR #223, opened 2026-07-10
Add dashboard subscriptions and alerts to integration setup wizard

MattPua:
## What

After the integration wizard builds a dashboard, it now offers to set up a **weekly email subscription** for that dashboard plus **up to two alerts** on its highest-signal insights — asking the user's consent first.

## Why

A dashboard nobody revisits is dead weight. This puts the wizard's output in an inbox and flags the metrics that actually matter, without wiring up noise.

Docs-only change to `context/skills/integration/references/4-conclude.md`.

---


gewenyu99:
@MattPua Run `/wizard-ci basic-integration` next time. all is... broken which is our fault haha

gewenyu99:
Haven't forgotten, trying to slot this in among the other things we're working on

PR #224, opened 2026-07-10
fix(self-driving): correct stale MCP tool names in read-context step

gewenyu99:
Step 2 named `survey-list` and `error-issue-list` (no longer exposed by the MCP server), costing the agent its first turn on each; switched to the current `surveys-get-all` and `query-error-tracking-issues-list`.


sarahxsanders (approved):
$$$$$

PR #237, opened 2026-07-14
feat(self-driving): connect credential inbox sources via the secure connect link

Gilbert09:
## What

Two related changes to the self-driving setup skill's connected-tool step:

1. **Add Jira** as a connected-tool inbox source (multi-select option, `jira`/`issue` responder mapping, already-connected detection).
2. **Connect credential sources for real, instead of arming dormant responders.** Zendesk, pganalyze, and Jira authenticate with API credentials, and the flow previously refused to collect them in-run, so it just enabled a *dormant* responder (a signal toggle with no warehouse source behind it) and left a follow-up. The responder stayed silent until the user separately went and connected the source.

## The fix

Reuse the existing **connect-link handoff** that already backs agent-driven source setup. For any API-credential tool the user picks, the self-driving run now:

- calls `data-warehouse-source-connect-link` to get a secure PostHog connect URL,
- hands the user that link (they enter credentials on a PostHog page in their browser, never in the chat),
- fetches the stored credential once via `data-warehouse-stored-credentials-list`,
- creates the **live** source via `data-warehouse-source-setup` with `{ "credential_id": ... }`.

Dormant responder + follow-up is now only the fallback when the user skips or can't finish, the same posture the Linear connector already uses when its OAuth grant doesn't land.

## Changes

- New connector `references/5c-credentials.md` (mirrors `5b-linear.md`, but the "one click" is entering credentials on the connect page).
- `references/5-connected-tools.md` — dispatch credential sources to the new connector; reframe the intro and class-recording language.
- `description.md` — the per-source connector confirms (now including the credential connect links) can't be batched.

## Context

No backend or wizard code needed. The connect object model (`PendingSourceCredential`), the connect page (`/data-warehouse/connect?kind=X`), and all three MCP tools already exist and ship today. This only teaches the self-driving flow to use them. Verified the skill bundle rebuilds and carries the connect-link flow.


Gilbert09:
Correction: this PR is merged (I misread it as sync-stuck). Follow-up #238 builds the connect-link handoff on top of it.

PR #241, opened 2026-07-15
fix(skills): notebook JSON escaping + report overwrite guidance (terra telemetry)

gewenyu99:
## What

Two prompt fixes in the `integration-v2` report step, driven by wizard remark telemetry from the gpt-5.6-terra rollout (2026-07-15, ~10 remarks in the first hour of external traffic):

1. **`skills/integration-v2/notebook`** — agents repeatedly failed `notebooks-create` calls by pasting raw multi-line report markdown into the JSON payload ("repeated notebooks-create JSON parsing failures"). The skill now:
   - says to `Read` the finished `posthog-setup-report.md` first (and not before the report step wrote it)
   - names the exact tool (`notebooks-create` via `posthog_exec`) so no tool-search round-trip is needed — one run wasted a lookup on a scope-gated alias surfaced first
   - spells out the escaping contract: `\n` / `\"` / `\\`, one single-line string, and fix-the-escaping-don't-trim-the-report on rejection

2. **`skills/integration-v2/report`** — remarks disagreed about read-before-write because the correct behavior differs: fresh runs must write directly (a read errors on the missing file), re-runs must `Read` first (harnesses refuse to overwrite an unread file). The skill now states both cases.

## Why now

All terra runs still succeeded — these loops cost retries, tokens, and ~seconds of wall-clock per run, not outcomes. But they're the top remark cluster by far (notebook escaping alone is 10+ mentions) and the fix is decoupled from the wizard release cycle.

## Notes

- `npm test` passes (121/121). `npm run build` is broken on the `experiment/orchestrator` tip **before** this change: `agents/integration-v2/build.md` (and possibly siblings) still declare `flow: "posthog-integration"` after the rename in 23276d8. Not addressed here to avoid colliding with in-flight rename work — flagging for whoever owns it.

---

### DIFF
--- a/context/skills/integration-v2/notebook/description.md
+++ b/context/skills/integration-v2/notebook/description.md
@@ -4,9 +4,13 @@ Once `posthog-setup-report.md` exists, mirror it into a shareable PostHog notebo
 so the user has an in-app copy to link and comment on. The notebook is an extra
 copy, not a replacement — keep the local report file in place.
 
-Call `notebooks-create` through `posthog_exec` with a `title` (e.g.
-`PostHog setup (wizard) – <repo name>`) and `content` set to a single markdown
-node that wraps the report verbatim:
+First `Read` the finished `posthog-setup-report.md` — don't reconstruct the
+report from memory, and don't try to read it before the report step has written
+it.
+
+Then call `notebooks-create` through `posthog_exec` — that exact tool name, no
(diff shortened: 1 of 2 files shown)


gewenyu99:
Superseded: notebook JSON-escaping reshaped as the integration-v2 notebook microskill, and dashboard split into dashboard+insight microskills (experiment/orchestrator). Closing this prompt-patch shape.

PR #243, opened 2026-07-15
feat(build): ship step-skill variants as one bundle per group

gewenyu99:
The four integration-v2 step groups expanded to 148 near-identical zips, so `packaging: bundle` now ships each group as one JSON holding every variant and the release carries 144 artifacts instead of 288.

### DIFF
--- a/scripts/lib/build-phases.js
+++ b/scripts/lib/build-phases.js
@@ -351,7 +351,13 @@ function writeBundles({ skills, sourceDir, skillsDir, merge = false, log = () =>
                 ? JSON.parse(fs.readFileSync(file, 'utf8')).variants
                 : {};
         const bundle = { id: group, variants: { ...existing } };
+        const seen = new Set();
         for (const skill of variants) {
+            // Two variants resolving to one short id would silently overwrite each other.
+            if (seen.has(skill.shortId)) {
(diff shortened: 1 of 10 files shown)


scripts/lib/build-phases.js
```diff
@@ -295,11 +315,61 @@ function validateDefault(entries) {
...
+    const artifacts = {};
+    for (const [group, variants] of Object.entries(groups)) {
+        const file = path.join(skillsDir, `${group}.json`);
+        const existing =
+            merge && fs.existsSync(file)
+                ? JSON.parse(fs.readFileSync(file, 'utf8')).variants
+                : {};
+        const bundle = { id: group, variants: { ...existing } };
+        for (const skill of variants) {
+            bundle.variants[skill.shortId] = readSkillFiles(path.join(sourceDir, skill.id));
```
sarahxsanders, line 361: should we validate IDs per bundle? I think if any duplicates slip through they could overwrite

sarahxsanders (approved):
are there any noticeable performance differences with JSON vs. zip with file sizes being larger? more-so curious

scripts/lib/build-phases.js
gewenyu99: Oh good call, there should be

PR #250, opened 2026-07-16
feat(scripts): visual-dags renders each flow's task graph

gewenyu99:
Adds `pnpm visual-dags`, which reads every agent prompt's frontmatter and renders each flow's `dependsOn` graph as a mermaid diagram in the browser, annotated with the model and effort each task runs at.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "build": "node scripts/build.js",
     "dev": "node scripts/dev-server.js",
+    "visual-dags": "node scripts/visual-dags.js",
     "test:plugins": "vitest run scripts/plugins/tests",
     "test:plugins:watch": "vitest scripts/plugins/tests",
     "test:skills": "vitest run scripts/lib/tests",
--- a/scripts/visual-dags.js
+++ b/scripts/visual-dags.js
@@ -3,10 +3,10 @@
 /**
  * Visualize agent DAGs
  *
- * Renders each flow's task graph — the `dependsOn` edges between its agent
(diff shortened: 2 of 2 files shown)


gewenyu99:
What `pnpm visual-dags` renders today — the whole `integration-v2` graph, each node annotated with the model and effort that task runs at, and the seed called out separately since it plans the graph rather than sitting in it.

integration-v2 agent DAG

Reading it top to bottom: `install` and `init` are independent roots, `identify` and `error-tracking` run in parallel off them, `capture` waits on `identify`, and `build` fans out to `dashboard` and `review` with `report` last.

gewenyu99:
Reworked — the first render was node-to-node edges, which buried the thing that actually matters. Now it groups each flow into the **waves the executor runs it in**: everything inside a box becomes runnable together, and the boxes stack in the order they run.

integration-v2 agent waves

Dropped mermaid to get here. It ignores `direction LR` inside a subgraph that participates in edges, so "2 run in parallel" rendered stacked vertically — reading as sequential, which is exactly backwards. Plain HTML/CSS gives the layout directly and removes the CDN dependency.

Worth noting what it makes obvious: `integration-v2` is 10 agents in 6 waves (2 → 2 → 1 → 1 → 2 → 1), and `capture` now occupies a wave by itself.

edwinyjlim (approved):
YES!!! 

let's eventually move this into a git workflow that can produce READMEs for docs

edwinyjlim:
i need this now

gewenyu99:
LOL @edwinyjlim you merged into my branch haha

edwinyjlim:
🤣 guess i'll wait

PR #252, opened 2026-07-20
fix(mcp-analytics): hand off for blocked installs

sarahxsanders:
## What it changes

This is the instruction sheet the AI agent follows to add PostHog to a project. Before, a blocked package install ended the run with an `[ABORT]`. This PR changes that: a sandbox-blocked install is treated as expected, not a failure. The agent finishes all the code changes and emits a non-fatal `[MANUAL-STEP]` line carrying the exact install command, so the run ends as a **success with one step left**, not an error.

It removes the `manual install required` abort case and rewrites STEP 3 (install) and STEP 7 (verify) so the agent never bails on a blocked install and never fakes a green summary on a project whose packages aren't installed.

## Why it is a handoff, not an abort

The agent runs inside a sandbox, a locked-down space that can only write inside the project folder. That limit is deliberate, and it is a security measure. The agent is driven by an LLM, and an LLM can be steered by the text it reads, so the sandbox keeps a bad or hijacked run from touching anything outside the project you pointed it at.

In a monorepo the install has to write a shared `node_modules` that lives above the project folder, which is outside the sandbox, so it fails. We are not loosening that boundary. Letting installs run outside the sandbox would also let the install scripts that ship inside packages run outside it, and running unknown code with full access to your machine is exactly what the sandbox exists to prevent. So the install genuinely belongs to a human.

The old approach modeled that as an abort. But an abort is a hard stop, and the agent usually installs before it writes code, so it would bail before doing any work. Weaker models also can't reliably do "install fails, keep going anyway, then abort at the end." Making the blocked install a non-fatal `[MANUAL-STEP]` fixes both: the agent always finishes the code changes, and the leftover install is surfaced honestly as one remaining step.

## User experience

The wizard still can't tell you "configured!" when it isn't. It finishes everything the agent can do, then hands off honestly: instead of a false success or a red error, the user gets a green success with the exact install command as the one step left.

## Merge order

These three PRs are one fix in three parts. Merge in this order:

1. [wizard #920](https://github.com/PostHog/wizard/pull/920): base of the wizard stack, merge first.
2. [wizard #929](https://github.com/PostHog/wizard/pull/929): stacked on #920, merge after it.
3. [context-mill #252](https://github.com/PostHog/context-mill/pull/252): release after #929, so the skill and the wizard light up together.

**This PR is step 3.** It is inert until #929 is live, and #929 is inert until this is, so they go together.

Nothing breaks if they land out of order. #929 and #252 are each harmless on their own; they just do nothing until both are live.

### DIFF
--- a/context/skills/mcp-analytics/description.md
+++ b/context/skills/mcp-analytics/description.md
@@ -22,8 +22,9 @@ If anything blocks instrumentation, **always** emit exactly one `[ABORT] <reason
 - `[ABORT] no mcp server found` — an exhaustive search (see the guardrail above) found no MCP server in the project.
 - `[ABORT] unsupported language for mcp analytics` — the server is neither TypeScript/JavaScript nor Python.
 - `[ABORT] could not locate the server entry point` — MCP signals are present, but the place the server is constructed or where requests are dispatched couldn't be found to instrument.
-- `[ABORT] manual install required: <command>` — the code changes are in place, but the SDK packages could not be installed. This is commonly a monorepo, where the install has to write a hoisted `node_modules` the current process can't write and fails with a permission error (`EPERM`). Put the **exact** install command for this project after the colon (e.g. `yarn add @posthog/mcp posthog-node`, `pnpm add @posthog/mcp posthog-node`, or the Python equivalent) so the wizard can show the user the precise line to run.
 - `[ABORT] <short specific reason>` — anything else that blocks the run (e.g. no readable project, or no PostHog credentials and no MCP server connected to fetch them). Keep it short and specific so it's useful when aggregated across runs.
+
+A blocked SDK install is **not** an abort. It is an expected handoff (see STEP 3): finish the code changes and flag the install with a `[MANUAL-STEP]` line so the run still succeeds with one step left for the user.
 
 ## Instructions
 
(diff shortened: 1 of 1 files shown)


context/skills/mcp-analytics/description.md
```diff
@@ -22,6 +22,7 @@ If anything blocks instrumentation, **always** emit exactly one `[ABORT] <reason
 - `[ABORT] no mcp server found` — an exhaustive search (see the guardrail above) found no MCP server in the project.
 - `[ABORT] unsupported language for mcp analytics` — the server is neither TypeScript/JavaScript nor Python.
 - `[ABORT] could not locate the server entry point` — MCP signals are present, but the place the server is constructed or where requests are dispatched couldn't be found to instrument.
+- `[ABORT] manual install required: <command>` — the code changes are in place, but the SDK packages could not be installed. This is commonly a monorepo, where the install has to write a hoisted `node_modules` the current process can't write and fails with a permission error (`EPERM`). Put the **exact** install command for this project after the colon (e.g. `yarn add @posthog/mcp posthog-node`, `pnpm add @posthog/mcp posthog-node`, or the Python equivalent) so the wizard can show the user the precise line to run.
```
gewenyu99: I am concerned the agent will reach for this too early. It'll try install _first_ fail and come here directly. Agents (like sonnet tier) aren't good at complex reasoning where "complete code changes then fail this way" 🤔 Just my concerns

gewenyu99 (approved):
Good stuff, but I am mildly concerned by using the abort pattern. Maybe this should all be come form of end state instead of "abort" because I think the agent will abort as soon as it hits the problem instead of finishing the job best it can before aborting

context/skills/mcp-analytics/description.md
sarahxsanders: yeah this is actually something I need to pause on. this is a valid concern. abort isn't the right call here on second thought.. it should just be a non-fatal success w/ a one step left screen 🤔

sarahxsanders: okay this is a bigger fix... gonna rework this PR :)

PR #258, opened 2026-07-21
fix(integration): be explicit that insights are created before events ingest

gewenyu99:
## Summary

Agents running the wizard's dashboard/insight step sometimes shipped an **empty dashboard shell** when the freshly instrumented events hadn't been ingested yet — and the skill text let them: it says to build insights "from the events this integration instrumented", with nothing about the (universal for new integrations) case where those events have no data yet. Several external runs in a row skipped insight creation for exactly this reason, and each left the same end-of-run remark asking us to clarify "whether dashboards should omit insights until newly instrumented events appear."

This makes the intent explicit in both integration flows (`integration` linear + `integration-v2`): insights are definitions over event names, they are *expected* to render empty at creation and fill in as events arrive, and pre-ingestion emptiness is never a reason to skip or defer them.

Deliberately a wording fix, not a hard gate — models that skipped did so from a defensible reading of an ambiguous spec, and legitimately-empty cases still exist.

## Why

Multiple production runs delivered success with a dashboard containing zero insights, each flagging the same instruction ambiguity in its remark. One sentence of explicitness closes it without constraining agent judgment.

---

### DIFF
--- a/context/skills/integration-v2/dashboard/description.md
+++ b/context/skills/integration-v2/dashboard/description.md
@@ -18,10 +18,11 @@ returned `id` — the insight step attaches every insight to it via
 ```
 
 Then create the insights (see the insight step) and attach each one to this
-dashboard's `id`. The dashboard is not done until those insights are attached —
-an empty dashboard shell is not a success, even when no events have been
-ingested yet (fresh insights render empty and fill in as events arrive; that is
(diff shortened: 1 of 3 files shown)


sarahxsanders (approved):
silly robot

PR #245, opened 2026-07-16
feat(agents): review stage for integration-v2 — parallel to dashboard, fixes what fails review

gewenyu99:
It is really cheap (<30s) to run a quick audit over the changes made before shipping, and while dashboards and reports are written. This is a big win, it can significantly improve code quality.

### See giant benchmark below as a comment

Adds a fresh-context `review` task agent after `build` (parallel to dashboard/report — it refactors and kills bad changes, never what is captured), quality principles in `init`, and `posthog-best-practices` for `identify` and `review`; bench evidence diffs on the gewenyu99 forks (healthchecks #9–21, monica #9–14, firefly-iii #10–13).

PR #260, opened 2026-07-21
feat(commandments): missing PostHog config fails loudly in dev, no-ops in prod

gewenyu99:
The dev-mode error message told the reader to remove the error once configured — a wording YARA reads as an instruction to remove PostHog. Reworded to state the fact instead of the instruction.


gewenyu99:
Wrong target — the actual commandment lives on #245, fixing there instead.

PR #265, opened 2026-07-22
Use documented tracing_headers option in example apps

gewenyu99:
## Why

An audit of the example apps flagged that they configure PostHog with `__add_tracing_headers` — an internal, `__`-prefixed name — rather than the public `tracing_headers` option shown in the SDK docs (e.g. [React](https://posthog.com/docs/libraries/react.md), [Next.js](https://posthog.com/docs/libraries/next-js.md)). Several other examples linked frontend and backend sessions by hand instead of using that option.

## What

Align every example app on the documented `tracing_headers` init option, and make it actually work:

1. **Rename `__add_tracing_headers` → `tracing_headers`** (code + READMEs) in Django, Nuxt 3.6, Nuxt 4, and React Router 7.
2. **Bump the SDKs** so the option is recognized (see below).
3. **Convert the hand-rolled apps** — astro-hybrid, astro-ssr, tanstack-start read `posthog.get_session_id()` and set an `X-PostHog-Session-Id` header manually (session ID only). They now use `tracing_headers`, which auto-injects both `X-POSTHOG-SESSION-ID` and `X-POSTHOG-DISTINCT-ID`.
4. **Use `window.location.hostname`, not `window.location.host`** — the extension matches on hostname (no port), so the port-bearing form silently no-ops in dev.
5. **Actually set the option in nuxt-3.6** — its client plugin only claimed to in the README.

## Due diligence

- **The option is real and public.** In `@posthog/types`: `tracing_headers?: string[]`; `__add_tracing_headers?: string[]` carries `@deprecated Use tracing_headers instead. Kept for backwards compatibility.`
- **Introduced in posthog-js `1.380.0`** — PostHog/posthog-js [#3715](https://github.com/PostHog/posthog-js/pull/3715). Below that, the extension reads only `__add_tracing_headers`, so `tracing_headers` is a silent no-op. Pins now: react-router & nuxt-3.6 → posthog-js `^1.406.2`; nuxt-4 → `@posthog/nuxt` `^1.7.80`; tanstack-start → posthog-js `^1.406.2` (it had resolved 1.338.0 via `@posthog/react`). Astro apps load posthog-js from the CDN snippet (always latest).

## Verification (real browser, Playwright)

Each app was driven through its login flow in headless Chromium and the outgoing `/api/auth/login` request headers were captured.

**The bug this caught.** The extension matches the request URL's `hostname` (no port). With `tracing_headers: [window.location.host]` the configured value is `["localhost:4321"]` in dev, which never matches `"localhost"`:

| tracing_headers value | headers on the login request |
|---|---|
| `[window.location.host]` (port included) | **none** — `{}`, despite a live client session |
| `[window.location.hostname]` (the fix) | `X-POSTHOG-SESSION-ID`, `X-POSTHOG-DISTINCT-ID`, `X-POSTHOG-WINDOW-ID` |

**Per-app result — all send the headers after the fix:**

| App | Client SDK | Headers on `/api/auth/login` |
|---|---|---|
| astro-hybrid | CDN snippet | ✅ |
| astro-ssr | CDN snippet | ✅ |
| react-router-7-framework | posthog-js + `@posthog/react` | ✅ |
| nuxt-3.6 | posthog-js plugin | ✅ (after adding the option) |
| nuxt-4 | `@posthog/nuxt` | ✅ |
| tanstack-start | bundled `@posthog/react` | ✅ |

**End-to-end session join (astro-hybrid).** With the client on the real SDK and the server's `posthog-node` captured at the wire, one login produced:

```
Client → POST /api/auth/login
  X-POSTHOG-SESSION-ID  = 019f8be0-3c55-77ec-a802-8a587c2e0eee
  X-POSTHOG-DISTINCT-ID = 019f8be0-29a3-7132-8f73-1e39e579d87a

Server posthog-node emitted:
  { event: "server_login", distinct_id: "…-demo",
    $session_id: "019f8be0-3c55-77ec-a802-8a587c2e0eee", source: "api" }
```

The server event's `$session_id` is exactly the header the browser sent — the client and server events land in the same PostHog session purely because of `tracing_headers`. (Login + logged-in screenshots were also captured locally.)

`npm test` (137 passing) and `npm run build` are green.

---

### DIFF
--- a/example-apps/astro-hybrid/README.md
+++ b/example-apps/astro-hybrid/README.md
@@ -15,14 +15,14 @@ This shows how to:
 - Opt specific pages into SSR with `export const prerender = false`
 - Keep most pages static for performance
 - Track events from API routes using `posthog-node`
-- Pass session IDs from client to server for unified sessions
+- Link client and server sessions automatically with the `tracing_headers` option
 
 ## Features
 
 - **Hybrid rendering**: Static pages by default, SSR when needed
 - **API routes**: Server-side endpoints for auth and event tracking
 - **Dual tracking**: Events captured on both client and server
-- **Session continuity**: Session ID passed to server via headers
+- **Session continuity**: Session and distinct ID forwarded automatically via `tracing_headers`
 - **Product analytics**: Track login and burrito consideration events
 - **Error tracking**: Manual error capture sent to PostHog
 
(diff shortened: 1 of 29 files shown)


gewenyu99 (commented):
seems legit. will be testing these

sarahxsanders (approved):
🚗

PR #270, opened 2026-07-22
feat(integration): add Java (Spring Boot) as a full example-based variant

sarahxsanders:
Adds a **full example-based** Java integration variant (not docs-only), anchored on a real, compiling **Spring Boot** app at `example-apps/java-spring-boot` — the most common Java backend framework, called out specifically in review.

**What's here:**
- `example-apps/java-spring-boot/` — `posthog-server` client as a singleton `@Bean` configured from the environment, flushed + closed on shutdown via `destroyMethod="close"`. A controller identifies users, captures events, evaluates feature flags with `evaluateFlags`, and reports errors as an explicit event (the Java server SDK has no automatic exception capture — documented, not faked). `gradle compileJava` clean.
- `config.yaml`: new `java` full variant, tagged `[java, java-server]`.
- `commandments.yaml`: `java-server` block — keyed `java-server` (not `java`) so backend rules never fold into the Android skill, which carries the `java` tag.

Ships as a pair with the wizard detection PR **PostHog/wizard#948**.

### DIFF
--- a/context/commandments.yaml
+++ b/context/commandments.yaml
@@ -97,6 +97,17 @@ commandments:
     - Call `IdentifyAsync` for known users and put PII such as email in person properties, not in event properties
     - Use `CaptureException(exception, distinctId, properties, groups, flags)` for handled exceptions; automatic exception capture is not available in the .NET SDK yet
 
+  # Keyed `java-server` (not `java`) so these backend rules never fold into the Android skill, which carries the `java` tag. The Java variant is tagged [java, java-server].
+  java-server:
+    - posthog-server is the PostHog Java server SDK; add the `com.posthog:posthog-server` dependency via Gradle or Maven and import from `com.posthog.server`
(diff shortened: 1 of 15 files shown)


gewenyu99:
Full e2e run of this PR pair (wizard #948 + context-mill #270, both with main merged in) against [spring-projects/spring-petclinic](https://github.com/spring-projects/spring-petclinic).

**Set up, one honest conflict** — the runner has no JRE so the agent could not self-verify. Idiomatic Spring: `@Bean(destroyMethod = "close")` for shutdown flush, blank config disables cleanly, `posthog-server 2.7.0` added to both `pom.xml` and `build.gradle`, exception handler wired, null-safe capture helper.

[Dashboard](https://us.posthog.com/project/228144/dashboard/2016546) and notebook created through prod MCP. Compiles clean, but petclinic's own `spring-javaformat` gate fails on `PetController.java` — shipped unverified because the runner lacks a JRE.

Note: the context-mill branch as filed fails at the dashboard step — its agent prompts predate the `posthog_exec` tool grant on main; works once main is merged (currently blocked on unsigned commits).

<details>
<summary>All 25 TUI frames</summary>



















































</details>

PR #278, opened 2026-07-24
fix(integration-v2): CSP guidance — script-src is required even when bundling

gewenyu99:
Root cause of the recurring "adds connect-src, misses script-src" review finding (e.g. wizard-workbench#3112 react-router/saas-template): init/description.md's "Work within it" bullet said bundling posthog-js satisfies script-src. It doesn't — replay recorder, surveys, and the toolbar lazy-load from the assets CDN as scripts regardless of how the entry bundle ships. Agents followed the skill faithfully and produced partial CSPs; the review dimension then deferred to the init handoff's "handled". Sol-effort reviews only caught it by over-reading into posthog-best-practices/references/session-replay.md, where the correct fact lives.

Three edits: init teaches the full directive set with a copy-paste block (script-src / connect-src / worker-src), the init handoff names exactly which directives were touched, and the build/review dimension verifies directive-by-directive rather than trusting the handoff — converting an accidental sol-only catch into a deterministic check both arms pass.

---

### DIFF
--- a/context/agents/integration-v2/init.md
+++ b/context/agents/integration-v2/init.md
@@ -30,5 +30,7 @@ file, never hardcoded in source, and `.env.example` lists the key names (with
 placeholder values) so the next developer knows what to set. Your handoff names the
 files you changed, how the client is constructed in them, and how a call site
 reaches that client — every step after you has to reach the same one the same way.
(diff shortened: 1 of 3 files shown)


edwinyjlim (approved):
man i almost forgot how annoying CSPs and proxy stuff were

PR #280, opened 2026-07-25
Add WooCommerce, Shopify, and WordPress to data-warehouse-source skill

Ref34t:
Both are live, full-availability PostHog data warehouse sources already (confirmed against posthog.com/docs/cdp/sources/), just not referenced from this skill's docs_urls — the skill only knew about Postgres, MySQL, Snowflake, BigQuery, and Stripe.

Verified: npm test (137/137), npm run build (manifest still emits cleanly).

### DIFF
--- a/context/skills/data-warehouse-source/config.yaml
+++ b/context/skills/data-warehouse-source/config.yaml
@@ -15,3 +15,5 @@ variants:
       - https://posthog.com/docs/cdp/sources/snowflake.md
       - https://posthog.com/docs/cdp/sources/bigquery.md
       - https://posthog.com/docs/cdp/sources/stripe.md
+      - https://posthog.com/docs/cdp/sources/woocommerce.md
+      - https://posthog.com/docs/cdp/sources/shopify.md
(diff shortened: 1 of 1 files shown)


Ref34t:
Added WordPress in a second commit, same shape as the first.

`/docs/cdp/sources/wordpress` documents a connector that syncs posts, pages, comments, media, categories, tags, and users over the WP REST API, but the skill didn't list it — so an agent picking a source had no way to know it exists. Same gap as WooCommerce and Shopify, one level over.

`npm test` 137/137, and the build emits `references/wordpress.md` into `data-warehouse-source-setup.zip` alongside the others.

Retitled the PR to match.

Ref34t:
Hey! This one's been waiting a week, so just pinging to keep it on the radar. Same fix as #282, one level over – WooCommerce, Shopify, WordPress are all live sources but the skill had no idea they existed. Tests pass, all good. Happy to change whatever, just let me know!

Gilbert09 (approved):

PR #285, opened 2026-07-28
fix(build): look up example skip patterns by directory name

### LINKED ISSUE #284: Per-example skip patterns are ignored for skills with a single example
## Problem

`context/skip-patterns.yaml` has an `examples:` section for rules that apply to one example app. For most skills those rules never run.

The Laravel entry says to skip `bootstrap/cache`. The built skill still contains it:

```
references/EXAMPLE.md
  ## bootstrap/cache/services.php     20,352 bytes
```

That is Laravel's generated package-discovery cache. It is a machine-written list of class names, and it is 17% of the whole Laravel skill. For comparison, `references/COMMANDMENTS.md` in the same bundle is 1,618 bytes.

## Cause

`scripts/lib/skill-generator.js:644`

```js
skipPatterns: mergeSkipPatterns(
  skipPatterns.global,
  skipPatterns.examples[isSingle ? skill.id : dirName]
)
```

When a skill has one example path, the lookup key is `skill.id`. Laravel's id is `integration-laravel`, but the key in `skip-patterns.yaml` is `laravel`, which is the directory name. No match, so the override is dropped and no error is raised.

When a skill has two or more example paths, the key is `dirName` and the override works.

## Two skills behave differently for this reason

- **Laravel** — one example path. Override ignored. `bootstrap/cache/services.php` ships.
- **Swift** — two example paths (`swift`, `swift-xcodegen`). Override applied. The `allow` for `project.yml` works and the file is correctly kept.

Same feature, opposite result, decided by how many examples the skill happens to have.

## Also affected

`android` has one example path, so its overrides are dropped too. Impact is smaller: `app/proguard-rules.pro` and `local.properties.example`, both small.

## Suggested fix

Key the lookup on the directory name in both branches, or look up `dirName` and fall back to `skill.id`.

Worth deciding deliberately rather than just patching, for two reasons. It changes what ships in existing bundles. And the two keying styles currently in the file (`laravel` and `swift-xcodegen` are directories, but a reader could reasonably assume skill ids) suggest the intended convention should be written down.

Separately, `bootstrap/cache/*.php` is generated output that Laravel's own `.gitignore` excludes, and it is committed to `example-apps/laravel`. Removing it would fix the Laravel symptom whatever is decided here.

Ref34t:
Fixes #284.

## Problem

Per-example overrides in `context/skip-patterns.yaml` were looked up by the wrong key for any skill with a single example, so they never applied.

`scripts/lib/skill-generator.js:644`

```js
skipPatterns.examples[isSingle ? skill.id : dirName]
```

Keys in the `examples:` block are example directory names — `laravel`, `android`, `swift-xcodegen`. With one example path the lookup used `skill.id` (`integration-laravel`) instead, missed, and returned `undefined`. `mergeSkipPatterns` treats that as "no overrides" and the build carried on with no error.

`isSingle` is the right switch for the other two things it controls — the output filename and the display title, both generated by the build. Config keys are typed by hand, and they are always directory names, so the same branch shouldn't apply here.

## Fix

One lookup, always by directory name, extracted so it can be tested:

```js
function skipPatternsForExample(skipPatterns, dirName) {
    return mergeSkipPatterns(skipPatterns.global, skipPatterns.examples?.[dirName]);
}
```

No key in `skip-patterns.yaml` is a skill id today, so nothing that currently resolves stops resolving.

## Effect on shipped skills

Two skills had overrides that were being dropped. Rebuilt and compared `references/EXAMPLE.md`:

| Skill | Before | After | Removed |
|---|---|---|---|
| `integration-laravel` | 81,558 | 60,662 | 20,896 (26%) |
| `integration-android` | 67,047 | 50,906 | 16,141 (24%) |

The Laravel case is almost entirely `bootstrap/cache/services.php`, Laravel's generated package-discovery cache — a machine-written list of class names that Laravel's own `.gitignore` excludes. It was 17% of that skill. For scale, `references/COMMANDMENTS.md` in the same bundle is 1,618 bytes.

`integration-swift` has two example paths, so it was already taking the `dirName` branch. Its `allow` rescue for `project.yml` still works after the change — verified, still 8 occurrences in `EXAMPLE-swift-xcodegen.md`.

## Tests

Five added to `scripts/lib/tests/skip-patterns.test.js`, including the one that pins the bug:

```js
it('does not look overrides up by skill id', () => {
    const patterns = skipPatternsForExample(config, 'integration-laravel');
    expect(shouldSkip('bootstrap/cache/services.php', patterns)).toBe(false);
});
```

The existing suite covered `mergeSkipPatterns` and `shouldSkip`, which were both correct — the bug was in the key handed to them, which nothing exercised.

```
npm test       ✅ 142 passing
npm run build  ✅ 118 skills
```

## Worth a follow-up

The lookup failed silently. A key in `examples:` that matches no example directory produces no warning, so a rule can look right in the file and do nothing. A build-time check that every `examples:` key resolves would turn this class of bug into an error. Left out here to keep the change to one thing — happy to add it if you want it.

Separately, `bootstrap/cache/*.php` is generated output committed to `example-apps/laravel`. This PR stops it reaching the bundle, but removing it from the repo would be worth doing on its own.

### DIFF
--- a/scripts/lib/example-processor.js
+++ b/scripts/lib/example-processor.js
@@ -75,6 +75,14 @@ function mergeSkipPatterns(globalPatterns, examplePatterns = {}) {
 }
 
 /**
+ * Global patterns plus any override for this example. The `examples:` block in
+ * skip-patterns.yaml is keyed by directory name, not skill id.
+ */
+function skipPatternsForExample(skipPatterns, dirName) {
+    return mergeSkipPatterns(skipPatterns.global, skipPatterns.examples?.[dirName]);
+}
+
+/**
  * Recursively collect all files in a directory
  */
 function collectFiles(dirPath, baseDir, skipPatterns) {
@@ -199,6 +207,7 @@ const defaultPlugins = [ignoreFilePlugin, ignoreBlockPlugin, ignoreLinePlugin];
 export {
     loadSkipPatterns,
     mergeSkipPatterns,
+    skipPatternsForExample,
     shouldSkip,
     processExample,
     defaultPlugins,
--- a/scripts/lib/skill-generator.js
+++ b/scripts/lib/skill-generator.js
@@ -41,7 +41,7 @@ import path from 'path';
 import crypto from 'crypto';
 import yaml from 'js-yaml';
 import matter from 'gray-matter';
-import { processExample, loadSkipPatterns, mergeSkipPatterns, defaultPlugins } from './example-processor.js';
+import { processExample, loadSkipPatterns, skipPatternsForExample, defaultPlugins } from './example-processor.js';
 import { CLI_ROLES, validateCommandName } from './cli-block-validation.js';
 
 /**
@@ -641,7 +641,7 @@ async function generateSkill({
                 displayName: isSingle ? skill.display_name : dirName,
                 id: skill.id,
                 repoRoot,
-                skipPatterns: mergeSkipPatterns(skipPatterns.global, skipPatterns.examples[isSingle ? skill.id : dirName]),
+                skipPatterns: skipPatternsForExample(skipPatterns, dirName),
                 plugins: defaultPlugins,
             });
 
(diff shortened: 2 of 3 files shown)


Ref34t:
Hey team! Just bumping this – it fixes the skip-pattern lookup bug so Laravel stops shipping 20KB of bootstrap cache noise (that's ~17% of the skill, ouch). Tests are in, everything passes. Would love a review whenever you've got a minute!

PR #254, opened 2026-07-20
feat(data-warehouse): offer 30 more sources in self-driving onboarding

Gilbert09:
## Summary

Make the `npx @posthog/wizard self-driving` onboarding offer the 30 new self-driving inbox
sources (support desks, issue trackers, error tracking, security scanners, product feedback
and reviews) that were added to the signals scout (PostHog/posthog#72378) and the Code inbox
UI (PostHog/code#3597).

The wizard's connected-tools list is hardcoded in the `self-driving` skill, so a new source is
invisible to onboarding until it's added here.

## Changes

In `context/skills/self-driving/references/5-connected-tools.md`:
- Added the 30 sources to the step-5 `wizard_ask` multi-select options.
- Extended the "already connected" warehouse `source_type` detection list.
- Extended the responder mapping (`source_product` / `source_type`) for each.

Credential sources reuse the existing generic connect-link connector (`5c-credentials.md`) —
no new per-source connector files. Also refreshed the source-list mentions in `description.md`
and `2-read-context.md`.

## Testing

- `node scripts/build.js` — the skill bundles and structure validation passes
  (self-driving-setup packaged, 115 skills).

PR #291, opened 2026-07-29
feat(self-driving): show codebase-detected tools first in the connected-tools ask

Gilbert09:
## Problem

The step-5 connected-tools multi-select grew from 5 → ~36 options ([#254](https://github.com/PostHog/context-mill/pull/254)) and overflows the terminal — you can't see the whole list ([thread](https://posthog.slack.com/archives/C09GTQY5RLZ/p1784927815366779)). The fix agreed there: show the tools detected in the user's codebase first, then the SaaS basics, then an "others" bucket.

## Changes

Restructures only the step-5 **ask construction** — the connect / enable / record machinery (steps 2–5) is untouched:

- **Detected first.** The wizard now injects a **"Tools detected in this codebase"** block (a deterministic dependency + env-key scan — companion PR [PostHog/wizard#1022](https://github.com/PostHog/wizard/pull/1022)). Step 1 puts the detected tools that match the catalog first.
- **Then the SaaS basics** — GitHub Issues, Linear, Jira, Sentry, Zendesk, always offered.
- **Then "Something else…"** — one option that expands to the full ~36-tool catalog in a second ask (step 1b), only when the user asks for it.

Default view drops from ~36 options to ~6–10.

### Degradation

Safe without the wizard PR: with no detected block, the first ask is just the basics + "Something else…" (still a big improvement), and ordering falls back to the old step-2 evidence. So the two PRs are independent — either can land first.

## Test plan

`pnpm build` (bundles the `self-driving-setup` skill) and `pnpm test:skills` both pass (285 skill tests). The change is a markdown reorder within one skill reference; no connector, source_type, or responder mapping changed.

*(Note: `pnpm security-scan:skills` fails locally with an unrelated `ERR_MODULE_NOT_FOUND` in the scan tooling — an env gap on my machine, not this change. Worth a CI check.)*

### DIFF
--- a/context/skills/self-driving/references/5-connected-tools.md
+++ b/context/skills/self-driving/references/5-connected-tools.md
@@ -29,12 +29,39 @@ Load `wizard_ask` via `ToolSearch select:mcp__wizard-tools__wizard_ask`. Reach `
 
 ## Do
 
-1. Ask **once**, multi-select. **"None of these" is the first option** (the safe default — an accidental `enter` declines); order the *tools* after it, seeding with any step-2 hints so a tool you saw evidence of comes first among them:
+1. **Build a short list from the codebase scan, then ask once.** The full catalog is ~36 tools — too many to show at once. The run prompt carries a **"Tools detected in this codebase"** block (a deterministic dependency + env-key scan the wizard ran on this project). Use it to keep the first ask short:
+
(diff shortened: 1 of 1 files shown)


gewenyu99 (approved):
Good stuff

PR #294, opened 2026-07-29
feat(feature-flags): add an end-to-end setup workflow

hacktivist123:
Turn the existing feature-flags skills into an ordered workflow that can assess an app, configure a safe flag, implement one use case, verify the result, and leave an evidence-based report.

## What

- Add a five-step workflow shared by the feature-flags skill variants.
- Add framework-specific browser and server rule tags for JavaScript variants.
- Update Next.js server evaluation guidance to use the current `evaluateFlags()` API.
- Separate public project tokens from personal API keys, OAuth tokens, and other secrets.
- Add conservative flag creation, rollout, verification, and reporting rules.
- Add a generation contract test for the Next.js skill.

## Why

The previous skill provided documentation and general principles, but it did not define an executable workflow or completion criteria.

Testing the generated Next.js skill with a real Wizard agent exposed gaps around environment-file detection, production-build verification, report ownership, safe rollout behavior, and credential terminology. These instructions turn those findings into reusable context rather than fixture-specific fixes.

## Verification

- `npm test` - 19 test files and 139 tests passed
- `npm run build` - all 363 skills generated
- Environment-variable naming check passed
- Generated Next.js skill inspected
- Bundle and individual skill archives remain within CI size limits
- End-to-end Wizard agent test completed against a Next.js App Router fixture and a live PostHog project
  - Production build and PostHog flag read-back passed
  - An existing inactive flag was reused without changing its rollout conditions
  - Environment configuration was checked by key presence without reading values

## Runtime coverage

The agent test covered the Next.js variant. Other feature-flags variants generated successfully but were not exercised through full runtime integrations.

Live evaluation and custom event ingestion were not run because the shared rollout was not changed solely for testing. The workflow requires these layers to remain visibly classified as `not run` rather than treating source review or local state as ingestion evidence.

### DIFF
--- a/context/skills/feature-flags/description.md
+++ b/context/skills/feature-flags/description.md
@@ -1,23 +1,29 @@
 # PostHog feature flags for {display_name}
 
-This skill helps you add PostHog feature flags to {display_name} applications.
+This skill adds a real PostHog feature flag to an existing {display_name} application. It inspects the application, finds or creates the flag, implements one safe use case, verifies the result, and leaves a short report.
+
+## Workflow
+
+{workflow}
+
+Start with `references/1-assess.md` and follow the steps in order. Each step links to the next one. Do not skip directly to implementation or preload every reference.
 
 ## Reference files
 
 {references}
 
-Consult the documentation for API details and framework-specific patterns.
+The workflow references control the job. Consult the framework documentation and `COMMANDMENTS.md` for API details and implementation patterns when a step asks for them.
 
 ## Key principles
 
-- **Environment variables**: Always use environment variables for PostHog keys. Never hardcode them.
-- **Minimal changes**: Add feature flag code alongside existing logic. Don't replace or restructure existing code.
-- **Boolean flags first**: Default to boolean flag checks unless the user specifically asks for multivariate flags.
-- **Server-side when possible**: Prefer server-side flag evaluation to avoid UI flicker.
-
-## PostHog MCP tools
-
-Check if a PostHog MCP server is connected. If available, look for tools related to feature flag management (creating, listing, updating, deleting flags). Use these tools to manage flags directly in PostHog rather than requiring the user to do it manually in the dashboard.
+- **Existing integration**: This workflow requires a working PostHog SDK integration. It does not install PostHog from scratch.
+- **Idempotent flag management**: Reuse an exact flag key. Never create a duplicate or silently change an existing flag's conditions, rollout, or evaluation context.
+- **Conservative rollout**: Do not expose a new experience broadly without the user's explicit intent.
+- **Minimal changes**: Add the flag alongside existing logic without restructuring unrelated code.
+- **Boolean flags first**: Use a boolean flag unless the requested behavior needs variants or a payload.
+- **Safe evaluation**: Resolve identity first, handle loading and failure states explicitly, and keep authorization or other security decisions outside feature flags.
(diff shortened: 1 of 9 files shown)


context/skills/feature-flags/references/1-assess.md
```diff
@@ -0,0 +1,62 @@
...
+
+```text
+[ABORT] The selected feature flag skill does not match this application. Rerun the command with the detected framework skill.
+```
+
+## Choose the use case
+
+Use a behavior named by the user when one exists. Otherwise, select one small, reversible user-facing enhancement supported by the code you inspected, and announce the choice before editing.
+
+Avoid authentication, authorization, billing, data deletion, privacy controls, and other behavior where a failed flag evaluation could create a security or data-integrity problem. A feature flag may control product behavior, but it must never be the security boundary.
```
edwinyjlim: really like the advice to avoid business critical areas. 

out of curiosity, what use cases did the wizard pick when you tested it? buttons? copy?

gewenyu99 (commented):
General feedback here for you is, what do you think the ideal use case is for this skill and wizard command?

What is the goal someone running this will accomplish when you designed this?

I think Feature Flags are unique in that they reflect a need, a requirement for someone to safely introduce new code or run an experiment.

Where should this intent of the user come into play here?

Do you think this approach will be meaningful for a potential feature flags user?

context/skills/feature-flags/description.md
```diff
@@ -1,23 +1,29 @@
...
-- **Minimal changes**: Add feature flag code alongside existing logic. Don't replace or restructure existing code.
-- **Boolean flags first**: Default to boolean flag checks unless the user specifically asks for multivariate flags.
-- **Server-side when possible**: Prefer server-side flag evaluation to avoid UI flicker.
-
-## PostHog MCP tools
-
-Check if a PostHog MCP server is connected. If available, look for tools related to feature flag management (creating, listing, updating, deleting flags). Use these tools to manage flags directly in PostHog rather than requiring the user to do it manually in the dashboard.
+- **Existing integration**: This workflow requires a working PostHog SDK integration. It does not install PostHog from scratch.
+- **Idempotent flag management**: Reuse an exact flag key. Never create a duplicate or silently change an existing flag's conditions, rollout, or evaluation context.
+- **Conservative rollout**: Do not expose a new experience broadly without the user's explicit intent.
```
gewenyu99: Not sure if it's related to this prompt or smwhere else, but I got a flag with 100% roll out, which is definitely not ideal.
Image

Not sure if this is intended or accurate, but the flag should be off with 0% roll out when introduced.

hacktivist123: That isn’t intended, and you’re right that it’s unsafe.
The current skill says to “prefer” 0% or an inactive flag, which leaves too much room for the tool’s default to win. I’ll make this a hard rule: create new flags with an explicit 0% rollout or inactive state, never rely on the API default, and read the flag back before continuing. If it comes back with a broad rollout, the agent must return it to 0% before changing application code or stop if it cannot do that safely.

context/skills/feature-flags/references/1-assess.md
hacktivist123: Thanks! In my Next.js todo tests which i got from the workbench, it picked display-only UI changes: a progress card showing total, active, and completed tasks, and a smaller completion summary. Both were reversible visual additions that didn’t affect authentication, billing, or data mutations etc.

PR #301, opened 2026-07-30
fix(wizard): Whack a mole part 8: exec-unavailable falls back to a subagent, never credentialed shell

gewenyu99:
State the fallback order in the shared `mcp-tool-calling` partial (inherited by every skill that calls PostHog tools): exec unavailable → delegate to a general-purpose subagent → skip the artifact and say so. Never the REST API with a credential in shell, source, or `.env`.

```
Before: exec refused → agent improvises `curl -H "Authorization: Bearer <secret>" …`
        → warlock: exfiltration_secret_via_shell, critical, blocked ×3 in one run (7cee73c4),
          then tries hardcoding a Personal API Key (reverted)
After:  exec refused → Agent-tool subagent runs the MCP call (subagents mount the PostHog MCP);
        if that fails too, the artifact is skipped and reported — no credential ever leaves the env
```

Companion to wizard [#1038](https://github.com/PostHog/wizard/pull/1038) (whack-a-mole 7), which fixes the underlying gate so exec is reachable from the main loop in the first place.

---


gewenyu99:
????

PR #302, opened 2026-07-30
feat(logs): add a logs-setup skill that correlates records to people and sessions

ThaliaBarrera:
## Problem

The `logs` skill group is docs-only. It declares twelve platform variants and points each at the relevant installation doc, but there's no step chain behind it — it's reference material, and nothing in the wizard drives it. Of the 238 `cliEntries` in a locally built `skill-menu.json`, none are logs-related.

Installation is the half those docs already cover well. The half they don't is **correlation**: attaching `posthogDistinctId` and `sessionId` to server-emitted records so a log line resolves to a person and opens their session replay. That's the reason to put logs in PostHog rather than in a standalone log vendor, and today it's left to the reader as something to implement by hand.

It's also the genuinely hard part, and a poor fit for a doc. The answer to *"where does this codebase already know who this request belongs to"* is different every time — an auth middleware here, a session callback there. A human reading docs has to work that out themselves. An agent reading the codebase can work it out and write the wiring.

## Changes

Adds `context/skills/logs-setup/`: a 7-step chain modelled on `migrate`, with `nextjs` and `python` variants.

`detect → install → plan → context → attach → verify → report`

Same conventions as `migrate` throughout: numbered `references/` with `next_step` frontmatter, one step in context at a time, state on disk between steps (`.posthog-logs-plan.md`), `[STATUS]` lines, `[ABORT]` cases, per-step `pass`/`warning`/`error`, and a `<wizard-report>` block. No commit — the operator reviews the diff.

Four decisions worth a reviewer's attention:

**Correlation attaches in exactly one place.** A `logging.Filter` on the OTel handler in Python; a logger wrapper in Next.js. The rule the skill enforces is that a codebase with two hundred log statements should come out of the run with two hundred correlated log statements and roughly one new file. Step 5 explicitly tells the agent to stop and move the shared mechanism if it finds itself editing a third call site by hand.

**The client half uses `tracing_headers`.** `posthog-js` already emits `X-POSTHOG-DISTINCT-ID` and `X-POSTHOG-SESSION-ID` from one config option, which is the convention `integration-v2/identify` and `posthog-best-practices` already document — the latter says explicitly "do not hand-roll a fetch wrapper for it". The public logs docs demonstrate threading `sessionId` through a request body instead, which means touching every route and missing anything that doesn't go through the wrapped function. The skill follows the in-repo convention and notes that these headers are client-controlled, so a server-known user id always outranks them.

**Correlation degrades in defined tiers.** `session` (both attributes), `person` (distinct id only), `none`. The report states which tier was actually reached and what would raise it. A background worker has no session and shouldn't pretend to have one. The failure mode this is written to avoid is wiring that looks correlated but attaches `undefined` — so Step 5 requires omitting an absent attribute entirely rather than passing an empty value, and requires that attachment never throws.

**The step files stay flat.** No `references/<variant-id>/` subdirectory, because `generateSkill()` filters to top-level `references/*.md` when copying into the zip. `migrate-statsig.zip` currently ships without the `references/statsig/` mapping docs its own steps point at. Happy to file that separately if it's news.

The skill also encodes three things the installation docs bury, all of which fail silently:
- the docs hardcode the **US** ingestion host, so Step 2 resolves the region from the project's existing PostHog host before writing an endpoint
- the export takes the `phc_` **project token**, not a `phx_` personal API key
- Next.js `middleware.ts` runs on the Edge runtime, a different runtime from the one the OTel provider is registered in, so request context established there is invisible to route handlers

## CLI surface

Ships as `role: skill`, reachable as `wizard skill logs-setup-<variant>`.

The companion wizard PR registers `wizard logs` as a native flat command, the same split `ai-observability` uses (`cli: role: skill` here, `nativeCommandFactory` there). Per CONTRIBUTING's promotion criteria I'd rather a maintainer decide whether this should also carry `role: command`; shipping as `skill` keeps it out of collision with the native registration in the meantime.

Companion PR: PostHog/wizard#1043.

## Test plan

- `pnpm test` — 137 passed
- `pnpm build` — builds `logs-setup-nextjs.zip` and `logs-setup-python.zip`; `skill-menu.json` goes 238 → 240 `cliEntries` with both new skills under a `logs-setup` category
- Verified the built zip is flat and complete: `SKILL.md`, all 7 steps, `COMMANDMENTS.md`, and the 8 fetched docs, with all `{display_name}` / `{references}` / `{commandments}` placeholders substituted and the `next_step` handoff footers wired 1→7
- `pnpm security-scan:skills` — zero findings in `logs-setup-*`. The scan reports 108 findings repo-wide; all are pre-existing in other skills and unchanged by this PR
- Served locally via `pnpm dev` and confirmed the wizard fetches both variants from `localhost:8765` with `--local-mcp`

## LLM context

Written with Cursor. The design came out of reading `migrate` as the structural template and `integration-v2/identify` + `posthog-best-practices` for the existing tracing-header convention, which is what redirected the client half away from the approach the public logs docs demonstrate.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/context/skills/logs-setup/config.yaml
+++ b/context/skills/logs-setup/config.yaml
@@ -1,0 +1,37 @@
+# PostHog Logs setup — installs the OTLP log export and correlates each record to
+# the session replay and person that produced it.
+#
+# Distinct from the `logs` skill group, which is docs-only reference material.
+# This one is a 7 step chain that edits the project.
+#
+# Reachable via `wizard skill logs-setup-<variant>`; the wizard also registers a
+# native `wizard logs` command that drives it.
+type: skill
+template: description.md
+category: logs-setup
+description: Set up PostHog Logs in an existing codebase and correlate each log record to the session replay and person that produced it. Additive — existing logging keeps working unchanged.
+tags: [logs]
+cli:
+  role: skill
+references:
+  preamble: "**Read ONLY this file.** Do not read any other reference file until this one tells you to."
+shared_docs:
+  - https://posthog.com/docs/logs/start-here.md
+  - https://posthog.com/docs/logs/link-session-replay.md
+  - https://posthog.com/docs/logs/link-person.md
+  - https://posthog.com/docs/logs/best-practices.md
+  - https://posthog.com/docs/logs/troubleshooting.md
+  - https://posthog.com/docs/logs/search.md
+variants:
+  - id: nextjs
+    display_name: Next.js
+    default: true
+    tags: [nextjs, javascript]
+    docs_urls:
+      - https://posthog.com/docs/logs/installation/nextjs.md
+
+  - id: python
+    display_name: Python
+    tags: [python]
+    docs_urls:
+      - https://posthog.com/docs/logs/installation/python.md
--- a/context/skills/logs-setup/description.md
+++ b/context/skills/logs-setup/description.md
@@ -1,0 +1,97 @@
+# PostHog Logs — {display_name}
+
+This skill sends this project's logs to PostHog and, more importantly, correlates them. A correlated log record carries the `posthogDistinctId` and `sessionId` of the request that produced it, so a log line in PostHog links back to the person who caused it and to the session replay of them causing it.
+
+Getting logs to arrive is table stakes and the installation docs already cover it. Correlation is the part that needs to read this codebase, and it is the reason this skill exists.
+
+## Workflow
+
+The setup runs as a 7 step chain.
+
+1. Detect the runtimes, the existing logging, and the PostHog SDKs already present (`references/1-detect.md`).
+2. Install the OTLP log export alongside the existing logging (`references/2-install.md`).
+3. Decide how far correlation can go in this codebase and persist the plan to `.posthog-logs-plan.md` (`references/3-plan.md`).
+4. Establish a request scoped identity context (`references/4-context.md`).
+5. Attach the identity to every log record from one place (`references/5-attach.md`).
+6. Verify the project still builds and a correlated record actually arrives (`references/6-verify.md`).
+7. Write the report and delete the plan file (`references/7-report.md`).
+
+Each step file points to the next via `next_step` frontmatter. Read them in order, one at a time. Do not preload future steps. Do not re-read a step file once you have moved past it.
+
+## Task list
+
+**Before you do anything else**, make a single call to `TaskCreate` to seed the task pane with the seven items above (one per step, in order). Use broad, job-oriented titles like "Detect logging setup", "Install log export", "Plan correlation", etc. The user is watching the pane and shouldn't see it sit empty.
+
+Then, and only then, read `references/1-detect.md` and start the chain.
+
+It's fine if your first list is imprecise. Call `TaskCreate` again (or `TaskUpdate` to refine existing items) every time your understanding sharpens. Use `TaskUpdate` to mark items `in_progress` when you start them and `completed` when you finish. Keeping the list current matters more than getting it right on the first call.
+
+## Correlation tiers
+
+Correlation is not all or nothing. Every project lands on one of these tiers, and the honest outcome is to reach the best tier this codebase supports and then say so.
+
+- **`session`** — the record carries both `posthogDistinctId` and `sessionId`. A log line links to the person and opens the session replay. This is the goal.
+- **`person`** — the record carries `posthogDistinctId` only. Logs appear on the person's profile, but there is no replay to open. This is the correct outcome when the project has no browser client, or when session replay is off.
+- **`none`** — no identity is reachable at log time. The logs still arrive and are still searchable, they just aren't linked to anyone.
+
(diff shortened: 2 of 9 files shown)


context/skills/logs-setup/references/1-detect.md
```diff
@@ -0,0 +1,61 @@
...
+
+Record roughly how many log call sites there are and which directories they cluster in. You do not need an exact count. You need to know whether this is a codebase with a logging convention or one with scattered `console.log`, because Step 5 attaches to the former and reports honestly about the latter.
+
+Note whether records are already structured. A logger that emits objects has somewhere to put correlation attributes. A logger emitting formatted strings does not, and Step 3 needs to know.
+
+## 3. Is PostHog already here?
+
+Search for `posthog-js`, `posthog-node`, `posthog` (Python), and any existing `posthog.init` or `Posthog(...)` construction. Note the host and token configuration, and where they come from.
+
+An existing PostHog host is the most reliable signal of the project's region, and Step 2 needs the region. Record the exact host value you found, and whether it came from a literal or an environment variable.
```
gewenyu99, line 45: You will have both region and project details from the Wizard to pass into the agent. I would rely on that and wire that information through.

context/skills/logs-setup/references/6-verify.md
```diff
@@ -0,0 +1,72 @@
...
+
+Step 2 added package entries without installing. Install once now, using the package manager the project already uses. Infer it from the lockfile, `pnpm-lock.yaml`, `package-lock.json`, `yarn.lock`, `poetry.lock`, `uv.lock`, or a `requirements.txt`.
+
+## Lint, type check, build
+
+Read the project's manifest for the scripts it defines. Run the linter and formatter only on the files you edited or created in this session, not across the whole codebase. Then run the project's type check (`tsc --noEmit`, `mypy`) and build (`next build`, `python -m py_compile`, or the project's own command).
+
+Capture stdout and stderr. Truncate to the failure region if the output is long.
+
+## Fix only what this run caused
```
gewenyu99, line 38: good scoping

gewenyu99 (commented):
The skills themselves look great! There are various nits but I'll stay very high level first.

Here's what I challenge you to think about. Logging is often very sensitive, it comes in at many different levels, it's often pushed across different logging providers, and so much of it is often intentionally anonymized to not get slapped with a GDPR lawsuit.

When should you identify logs? Should you decide for the user or suggest to the user? How do you think about compliance?

If there is a good reason not to log and associate with a person, what other reasons make it valuable to capture logs, especially to PostHog? Think about the context and tasks that push you to reach for logs.

I also suggest you think about logging for monorepos. This is something we haven't solved for the main program (though we're in progress there), but for logs this is especially important. Most logging happens at the service level in the backend for good reason. How do you think about this problem?

Some helpful reading material:
https://posthog.com/docs/logs/pii-scrubbing#quest-item-use-for-free

context/skills/logs-setup/config.yaml
```diff
@@ -0,0 +1,37 @@
...
+#
+# Reachable via `wizard skill logs-setup-<variant>`; the wizard also registers a
+# native `wizard logs` command that drives it.
+type: skill
+template: description.md
+category: logs-setup
+description: Set up PostHog Logs in an existing codebase and correlate each log record to the session replay and person that produced it. Additive — existing logging keeps working unchanged.
+tags: [logs]
+cli:
+  role: skill
```
sarahxsanders, line 15: this means the wizard program runs with the command `wizard skill logs-setup-nextjs` when your intention with the wizard PR looks like to have it run with `wizard logs`

we have some docs on posthog.com about wizard commands that may help here!

context/skills/logs-setup/description.md
```diff
@@ -0,0 +1,97 @@
...
+
+It's fine if your first list is imprecise. Call `TaskCreate` again (or `TaskUpdate` to refine existing items) every time your understanding sharpens. Use `TaskUpdate` to mark items `in_progress` when you start them and `completed` when you finish. Keeping the list current matters more than getting it right on the first call.
+
+## Correlation tiers
+
+Correlation is not all or nothing. Every project lands on one of these tiers, and the honest outcome is to reach the best tier this codebase supports and then say so.
+
+- **`session`** — the record carries both `posthogDistinctId` and `sessionId`. A log line links to the person and opens the session replay. This is the goal.
+- **`person`** — the record carries `posthogDistinctId` only. Logs appear on the person's profile, but there is no replay to open. This is the correct outcome when the project has no browser client, or when session replay is off.
+- **`none`** — no identity is reachable at log time. The logs still arrive and are still searchable, they just aren't linked to anyone.
```
sarahxsanders, line 35: this rocks! you're looking at it from what partial success looks like and thinking about those outcomes

sarahxsanders (commented):
I feel like this is the stronger of the pair of PRs! the correlation tiers are well thought through. a few things:
1. step 6 can't check what it claims which could report a false failure
2. the command here plus `wizard logs` in the other PR means there's two ways to run the same thing. we have docs on this if you go take a look! they should be easy to search for 
3. feel free to open issues for bugs you find!

PR #311, opened 2026-08-02
fix(docs): repoint example-app paths after the examples/basics move

Angiephoenix:
The example apps moved from `examples/basics/` to `example-apps/`, but the docs
still point to the old layout.

Nine example READMEs, laravel's IMPLEMENTATION.md, and the root README print `basics/<name>/` in their structure blocks, and expo's quick start says `cd basics/expo`, which
fails. These aren't just local docs. example-processor.js reads every file in an
example directory, README first, and that markdown ships inside the skill
bundles. `.md` isn't in skip-patterns.yaml, so
`grep -rn "basics/" --include="*.md" dist` returned hits across
dist/marketplace/plugins/... before this change. After it, nothing.

The root README tree was out of date in three ways: `examples/` doesn't exist,
the `basics/` level doesn't exist, and it listed 7 examples where there are now 40.

One extra: javascript-web's README pointed at `basics/javascript/`. The
directory is `javascript-web`, so the name needed updating along with the
parent.

Two things I'm less sure about. I left `mcp-commands/` out of the tree because I
couldn't find that directory anywhere in the repo. I also left out `scripts/`,
since it sits at the repo root rather than inside the examples. Happy to add
either back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,17 +37,47 @@ These are more like model airplanes. They're dramatically simplified to make it 
 But the leanness makes these useful for agent-driven development. Use these as context to help your agent make better integration decisions about PostHog.
 
 ```
-examples/
-├── basics/
-│   ├── next-app-router/         # Next.js 15 with App Router
-│   ├── next-pages-router/       # Next.js 15 with Pages Router
-│   ├── react-react-router/      # React with React Router
-│   ├── react-tanstack-router-file-based/   # React with TanStack Router (file-based)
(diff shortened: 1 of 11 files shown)


Angiephoenix:
Rebased on main and added the flutter example, so the tree matches the current
36. Still docs-only, and the twelve `basics/` paths are all still on main.

The two Semgrep checks are waiting on workflow approval, so they can't run
until a maintainer approves them. Happy to adjust anything.

gewenyu99 (approved):

PR #319, opened 2026-08-04
feat(ai-observability): rebuild the skill around the SDK wrapper and the session tree

marco-g-pm:
Rewrites the `ai-observability` skill to follow PostHog/posthog#75976, which moved the AI Observability installation docs off OpenTelemetry and onto the PostHog SDK wrappers.

## Why

The docs PR states the reason directly: *"OTel doesn't give the best developer experience to end up with a full session tree (session, traces, spans, generations). It makes it difficult to capture sessions in an easy to develop, maintain, and understand way."* That session tree is exactly what this skill exists to produce, so the skill had to follow.

Counting tokens across that diff shows how total the reversal is:

| | added | removed |
|---|---|---|
| `opentelemetry` | 4 | 207 |
| `PostHogSpanProcessor` | 0 | 92 |
| `TracerProvider` | 0 | 46 |
| `posthog_trace_id` | 9 | 0 |
| `$ai_span` | 69 | 3 |

`openai.tsx` ends up with zero OTel mentions, and all fifteen OpenAI-compatible gateways collapse into one shared snippet.

## What changed

The skill previously stopped at the bootstrap: a successful run captured individual `$ai_generation` events but never built the session → trace → span → generation tree. Running it twice against one unchanged fixture produced the full tree once and bootstrap-only the other time — both passing the skill's own bar. The steps are now:

- **`1-begin`** — variant picker reaches all 68 variants (frameworks before providers, `baseURL` check so gateways stop misrouting to `openai-*`). Asks what identifies the conversation, the user, and one turn.
- **`2-install`** — providers and gateways install just the PostHog SDK beside the vendor SDK. No OpenTelemetry packages.
- **`3-instrument`** (renamed from `3-otel-setup`) — wrapper client is the default mechanism. Copy the doc's shape: no init function, no module globals, no extra env scaffolding.
- **`4-nesting`** (new) — identity is per call; the trace id is per turn and shared across that turn's calls. Tool executions are captured as `$ai_span` events, since the wrapper never sees the app's dispatch loop.
- **`5-verify`** (renamed from `4-verify`) — grades what lands in PostHog rather than what the diff contains.

Two supporting fixes:

- **`commandments.yaml`** — carve-outs for rules that assume a `Posthog()` client this skill's paths don't create, and for the PII rule that contradicts `$ai_input` deliberately carrying the user's prompt.
- **`skill-generator.js`** — dedupe tags before collecting commandments. Variants re-declare group-level tags, which was doubling every rule. Covered by a test.

## Verified against source, not assumption

- **AWS Bedrock stays on OTel.** It was never part of #75976 and has no wrapper client, so it keeps `opentelemetry-instrumentation-botocore`. Found by diffing this skill's 43 variant frameworks against the 42 onboarding pages.
- **A raw provider SDK + OTel instrumentor emits no tool spans.** Read from the published `opentelemetry-instrumentation-anthropic` wheel: it patches only the vendor SDK's own methods, and `tracer.start_span` appears exactly twice in the package. Tool use is recorded as attributes on the generation, never as a child span. This is why manual `$ai_span` capture is now the documented answer.
- **Gateways must send `$ai_provider`.** `posthog.ai.openai` reports `openai` whatever host it targets, and PostHog prices by `$ai_model` + `$ai_provider` — see PostHog/posthog-python#822.


scripts/lib/skill-generator.js
```diff
@@ -298,7 +298,9 @@ function expandSkillGroups(config, configDir) {
         const compositeKeyDashed = key.replace(/\//g, '-');
 
         for (const variation of group.variants) {
-            const mergedTags = [...baseTags, ...(variation.tags || [])];
+            // Deduped: a variant that repeats a group-level tag would otherwise
+            // collect that tag's commandments twice.
+            const mergedTags = [...new Set([...baseTags, ...(variation.tags || [])])];
```
edwinyjlim, line 303: 🙏

context/skills/ai-observability/references/1-begin.md
```diff
@@ -1,52 +1,70 @@
...
-- Google Gemini: `genai.Client(`, `new GoogleGenerativeAI(`
+Language follows the manifest. A `package.json` means Node. A `pyproject.toml` or `requirements.txt` means Python. Framework variants have no language suffix.
 
-Note the app's entry point (server startup file, `main.py`, `index.ts`, `instrumentation.ts` in Next.js, etc.) — OTel must be initialized *before* the vendor SDK is imported, and the entry point is where that happens.
+Report the variant and the reason in a `[STATUS]` line, then call `install_skill` with the full id.
 
-Do not edit yet. Once you have a note of the entry point and the call sites, move on to `2-install.md`.
+## Read four facts from the code
+
+The install doc holds the code. It cannot know this app. Write these answers down. Step 3 uses them and nothing else.
```
edwinyjlim: does the agent actually write this to disk anywhere? we do offload context in other wizard programs this way, just curious since there's no specified scratch file

edwinyjlim (approved):
adding example code or explicitly calling out the documentation in the references might improve the behavior, but we'll cross that bridge after usage

let's good

context/skills/ai-observability/references/1-begin.md
marco-g-pm: oh I assumed this wrote to it's memory or something. is there something that I need to specific for it to write to a specific scratch file?

edwinyjlim: every model kinda interprets this differently. it's okay if you're not seeing unintended behavior, but "write" might actually create a file

when we want a persistent task list or notes, we have a combination of tool calls and prompts to update a temporary file we delete at the end of the run. sounds like you don't need that quite yet

edwinyjlim: I'd remove "write these answers down" just in case. It'll have access to the previous context with each turn

PR #323, opened 2026-08-04
fix(ci): sign the commits this workflow creates

Piccirello:
## Summary

Makes the workflow that commits in this repo produce **signed** commits.

Commits created with plain `git` are unsigned. PostHog is rolling the org-wide "Require signed commits" ruleset out to every repo, and this workflow would be rejected with `GH013` once it applies here. Routing the commit through the GitHub API instead means GitHub signs it with its own key, so it lands verified.

Found while inventorying which workflows across the org still create unsigned commits. `planetscale/ghcommit-action` is the pattern already in use in `posthog-js`, `posthog-roblox`, `brand`, `charts` and the SDK release fleet.

## Changes

The two mirror pushes (`PostHog/skills`, `PostHog/ai-plugin`) now commit through the API.

## Testing

Not run end to end. The working clones moved from `mktemp -d` into `$GITHUB_WORKSPACE/.skills-push` and `.ai-plugin-push`, because `ghcommit-action` resolves its `repository` input relative to the workspace. Both target repos already enforce signing, so these pushes were failing before this change.

---

## Review follow-up (pushed)

- Both mirror clones are now removed in an `if: always()` step. They embed the App token in `.git/config`, and moving them from `mktemp -d` into `$GITHUB_WORKSPACE` meant the token stayed readable by every later step in the job.

## Known limitation worth a decision before merge

**The API commit path cannot set an executable bit.** GraphQL `FileAddition` accepts only `path` and `contents` (verified by schema introspection), so a *newly* mirrored executable file lands as `100644`. Whether an *existing* `100755` file keeps its mode when its contents change is undocumented and I could not confirm it — no PostHog repo currently has an executable file that a `ghcommit` workflow has ever written.

Blast radius today is two files under the tree this workflow mirrors into `PostHog/skills`:

- `skills/posthog/all/hooks/skill-reminder.sh`
- `skills/team/customer-success/impersonation-toolkit/scripts/impersonate-audit.sh`

Both are currently `100755`, and both were last written by the old plain-`git` mirror. If either loses its exec bit, the hook silently stops running. Worth checking the mode on the first real mirror run after this merges.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -248,15 +248,23 @@ jobs:
           repositories: skills,ai-plugin
 
       - name: Push plugins to skills repo
+        id: skills_push
         env:
           SKILLS_REPO_TOKEN: ${{ steps.skills_token.outputs.token }}
         run: |
           MANIFEST="dist/push-manifest.json"
           TARGET_REPO=$(jq -r '.target_repo' "$MANIFEST")
-          WORK_DIR=$(mktemp -d)
+          # ghcommit-action resolves its `repository` input under $GITHUB_WORKSPACE,
+          # so the clone lives there rather than in a mktemp dir.
+          WORK_DIR="$GITHUB_WORKSPACE/.skills-push"
+          rm -rf "$WORK_DIR"
 
           echo "Cloning $TARGET_REPO..."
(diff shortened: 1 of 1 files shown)


edwinyjlim (approved):
tyty

PR #329, opened 2026-08-06
feat(audit-feature-flags): rebuild as a doctor (wizard native)

ryanwaits:
## Problem

Following up on the flags-doctor thread:

`audit feature-flags` is static and read-only today: greps the source, reads the roster, hands back a report. It never checks whether flags reach the running app, and can't act on what it finds. Its own report template signs off with "runtime assumptions unproven by the static code."

So a typo'd flag key returns `undefined` forever with no error, and since the audit only compares tenant→code, it can't see it. On a test app with one planted, the current skill told me to "confirm that every active flag matches one of the four keys referenced in code: `beta-serach`, …" — i.e. validate the typo as truth.

## Changes

Rebuilt as a doctor: verify (read-only) → one `wizard_ask` multi-select → apply only what's selected. Shaped after `web-analytics`'s doctor per Edwin's suggestion. Skill id, `cli:` block, and command unchanged. All nine static checks kept.

New:
- **Live delivery** — probes `/flags` with the project's key via the app's configured host, cross-checks the roster, surfaces per-flag `reason` codes.
- **Ghost keys** — code→tenant direction.
- **Evaluation reporting** — catches `$feature_flag_called` suppression, config-level and per-call.

**The interlock**, from Vincent's "evaluating ≠ sending events" point: staleness is measured from `$feature_flag_called`, so a heavily-used flag with suppressed events looks unused — archive it and you turn off a live feature. The doctor withholds all tenant-side cleanup until reporting is verified, and explains why.

Bot filtering of `/flags` is expected behavior, so it's a teaching note rather than a finding — and probes send a browser-like UA so we don't false-alarm on it.

## Test plan

- `npm test` (140) and `npm run build` green; zip builds, warlock clean, `cliEntries` unchanged.
- End-to-end on a Next.js app with three planted defects (`--local-mcp` + the wizard PR): all three caught with the right severities, two selected fixes applied surgically, archive candidate withheld by the interlock, healthy flags verified delivered.
- Same app on the published skill: all three silent.


ryanwaits:
Specific areas, if you're offering: mostly trade-offs I made deliberately but am not totally certain about yet:

**1. Speed vs. depth.** The old audit was static-only and quick. This one also probes `/flags`, fetches the roster, and asks for consent, so it's meaningfully slower. Read-only is still the default outcome (tick nothing, nothing happens; CI mode skips the fix phase entirely), but "I just want the fast static pass" is now a worse experience than it was. I've got `--report-only` and per-area flags (`--only delivery`) scoped as follow-ups.

**2. Rebuilding in place vs. a new skill.** I kept the id, command, and `cli:` block so there's no orphan and no second thing to maintain, but it does mean anyone who ran this expecting a read-only report now meets a prompt. Reasonable, or would you rather the doctor lived alongside the old one for a release?

**3. Check calibration.** Severities are my judgment, and across runs I saw one borderline check flip between `pass` and `suggestion`. If any thresholds feel wrong to you, that's the most valuable thing you could call out.

**4. Prompt weight.** More checks means a longer skill. It's comfortable on a small app but I haven't run it against a large monorepo...

context/skills/audit-feature-flags/description.md
```diff
@@ -1,74 +1,172 @@
...
+## Available tools
+
+{{> mcp-tool-calling}}
+
+**Verify (read-only):**
+- `Glob` / `Grep` / `Read` — static checks over the source tree.
+- `Bash` (plain `curl` only) — live `/flags` probes. Keep probe commands minimal and legible; see `references/checks.md` for the exact shapes.
+- `feature-flag-get-all` (or the equivalent MCP flag-listing tool; `execute-sql` fallback) — the project's flag roster.
+- `mcp__wizard-tools__check_env_keys` — which env keys exist (never reveals values).
+- `docs-search` — latest doc URLs for remediation links.
```
edwinyjlim, line 40: nice

context/skills/audit-feature-flags/config.yaml
```diff
@@ -1,20 +1,19 @@
...
 template: description.md
-description: Audit a PostHog integration's feature flag usage for correctness and cost-optimization opportunities
+description: Verify, diagnose, and fix a PostHog feature flags integration — live delivery checks plus correctness and cost audits, with consent-gated fixes
 tags: [best-practices]
 cli:
   role: command
   parentCommand: audit
   command: feature-flags
-references:
-  preamble: "**Read ONLY this file.** Do not read any other reference file until this one tells you to."
```
edwinyjlim, line 10: FYI this is a prompt guardrail for the agent to follow the references in a strict sequence. I know you've defined the flow in the `description.md` but sometimes agent ignores instructions

context/skills/audit-feature-flags/references/checks.md
```diff
@@ -0,0 +1,531 @@
...
+
+Every check is independent and required. A failure in one does not block the others. Do not invent checks beyond the ones listed.
+
+---
+
+## Part 1 — Static checks (parallel subagents)
+
+Emit `[STATUS] Auditing feature flag correctness`, then make **six `Agent` tool calls in a single message** for the correctness checks (Tasks A–E and Task I). When all five return, emit `[STATUS] Auditing feature flag cost optimization` and dispatch the cost checks (Tasks F–H) the same way — one `Agent` call per check that actually runs, in a single message. Two cost checks are gated on the pre-flight local-evaluation signal; for a gated check that is skipping, emit its `audit_resolve_checks` update directly (`status: "pass"`, `details: "skip: local evaluation not detected"`) instead of dispatching a subagent.
```
sarahxsanders: this contradicts a little, "six agent tool calls" - "When five return", looks like it's left over from the old version possibly, but I'd clean this up because I could see the agent tripping on this and wasting turns

context/skills/audit-feature-flags/references/checks.md
```diff
@@ -0,0 +1,531 @@
...
+Rule:
+- pass: no production-path suppression in code AND (query unavailable OR calls > 0 OR the project has no production traffic yet — do not fail a brand-new project on zero events alone; record `details: "no traffic baseline"`).
+- warning: code suppresses `$feature_flag_called` in production paths, OR the project has flag call sites and recent traffic but zero `$feature_flag_called` events in 7 days.
+- error: suppression is unconditional AND the roster shows flags used in experiments (roster `filters`/experiment linkage) — experiment exposure is silently broken.
+
+`details` (compact JSON): `{"suppression_sites": [{"file": "<path:line>"}], "calls_7d": <N or null>, "gates_cleanup": true | false}` — set `gates_cleanup: true` whenever status is warning/error; Phase 2 reads this to withhold tenant-side cleanup options.
+
+### Check N — `ff-stale-rolled-out` (sweep)
+
+What it verifies: flags at **100% rollout with no conditions** that are still gated in code — the check is equivalent to a hardcoded value: dead branches, needless evaluations, cleanup candidates. (The tenant-side twin, zero-reference active flags, is Task I in Part 1 — `ff-active-but-unreferenced`.)
```
sarahxsanders: I think there's some stuff here that needs cleaned up, is left over from the old version or maybe edits you did! you mention the wrong task and check

also side note: this is a superrrrrr long file, I wonder if the agent is a little slow based on the file side and sheer amount of context here?

context/skills/audit-feature-flags/references/report-format.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Doc links per check
+
+- `ff-bootstrap-when-known-set`, `ff-bootstrap-distinct-id-mismatch` → https://posthog.com/docs/feature-flags/bootstrapping
+- `ff-await-readiness`, `ff-default-values`, `ff-identified-only-pre-auth-targeting`, `ff-eval-before-identify` → https://posthog.com/docs/feature-flags/best-practices
+- `ff-key-authenticates`, `ff-flags-endpoint`, `delivered-*`, `ghost-*` → https://posthog.com/docs/feature-flags/troubleshooting
+- `ff-evaluated-not-reported` → https://posthog.com/docs/experiments/exposures
+- `stale-*`, `ff-active-but-unreferenced` → https://posthog.com/docs/feature-flags/cleaning-up-stale-flags
+- `ff-local-eval-polling-interval`, `ff-local-eval-in-edge-handlers` → https://posthog.com/docs/feature-flags/local-evaluation
+- `ff-test-ci-gating` → https://posthog.com/docs/feature-flags/cutting-costs
```
edwinyjlim, line 97: ooh mapping tasks to docs is interesting

context/skills/audit-feature-flags/description.md
```diff
@@ -1,74 +1,172 @@
...
+
+- If there are **no findings**, skip the confirm step, write a clean-bill report, and stop.
+- If findings exist but **none are fixable** (all `manual`), skip the confirm step, write the report with them under "Manual follow-up", and stop.
+- Otherwise call `mcp__wizard-tools__wizard_ask` **exactly once**:
+  - `kind: "multi"`.
+  - `prompt`: short summary, e.g. "Found N feature flag issues. Select the ones you'd like me to fix:". Mention manual-only findings are in the report.
+  - `options`: one `{ label, value }` per fixable finding. `value`: the check id (or `"<checkId>:<flag-key>"` for per-flag rows).
+  - **`label` is ONE short line — 80 characters or fewer, no newlines.** The option schema has no description field, so a multi-line label is a wall of text in the overlay. Format: `"[<SEVERITY>] <what will change> — <file or flag>"`. Reasoning, evidence, and file:line detail belong in the report, never here.
```
sarahxsanders: if you're using the anthropic harness, there is a `description` field (it's optional)

edwinyjlim (commented):
overall, this is a solid list.  the checks.md file is a little unwieldy though. i'd imagine it's going to be hard to scale prompts inline like that, but i like the task <> subagent structure

gewenyu99:
Some discussions based on what you raised. Feel free to discuss but defer execution if you have ideas. I again don't expect any of this to happen during this short superday:

1. Speed vs. depth. There are two things to know here: do you think the model and effort choices make sense right now? How might you approach optimizing from this perspective? We also have a DAG based way to create execution graphs of mini agents, how might this change your assumptions?

And if, say, we can't make it faster, what would you optimize for? speed vs. depth? how would you make this decision or run an experiment.

2. Do you think they should stay separate or should all audits become doctor skills? How might you think about architecting this. 

3. Check calibration: How would you calibrate this in the wild once this is released?

4. Prompt weight: how do you think about realistic upper limits of this wizard program? What are the real constraints, and what are subjective ones we have to judge based on taste/feedback?

sarahxsanders (commented):
this is a big step up from what it replaced, well done!! some high level stuff:
- going from grep the source to actually probing flags is great
- some contradictions in the context that need some cleaning up, I'd be curious to see how many turns the agent wastes when it gets tripped up by one!

context/skills/audit-feature-flags/config.yaml
ryanwaits: ok thats fair, good catch - ive restored as prose guardrails

context/skills/audit-feature-flags/references/checks.md
ryanwaits: definitely, good catch

context/skills/audit-feature-flags/references/checks.md
ryanwaits: yep, it was six after all - good catch again. i also also relettered to kill the task/check collision

context/skills/audit-feature-flags/description.md
ryanwaits: done! scoped claim to `wizard_ask` and added guidance to use `description` on harness native tools

ryanwaits:
@gewenyu99 on speed vs depth: i think i'd optimize for speed, because depth is something you can ask for specifically later. the checks already group into areas, so `--only delivery` or `--only observability` gets you the deep version of the thing you actually came here about. people already have the primitives to dig into any one of these (mcp,api, etc) they don't need the wizard to bundle all of it every run. the quick pass telling you where to look is still good enough in many ways

on separate vs all doctors: separate commands, same content, mostly so we don't end up with two copies of the check wording quietly drifting apart

PR #344, opened 2026-08-19
feat(mcp-analytics): detect and instrument MCP TypeScript SDK v2 servers

gesh:
`@posthog/mcp` has supported both MCP TypeScript SDK majors since 0.11.1, and PostHog/posthog.com#19545 just documented them. The `mcp-analytics` skill still described a v1-only world.

## The one that matters

STEP 1 listed `@modelcontextprotocol/sdk` as the only official-SDK detection signal, and its guardrail aborts with `[ABORT] no mcp server found` when nothing matches. A v2 project depends on `@modelcontextprotocol/server` and `/core` and has **no** `@modelcontextprotocol/sdk` at all — so `wizard mcp-analytics` could turn it away before opening a single reference doc.

That is why the docs PR alone could not fix v2 projects: detection gates whether the agent ever reaches the docs.

## Also corrected

| | |
|---|---|
| Path A | Named only v1's import. Now carries both, and notes v2 registers tools with `registerTool()` (`server.tool()` was removed). The `instrument()` call is unchanged — shape is detected at runtime, so emitted code must not branch on the major |
| STEP 3 | Silent on peers. Now: never install an MCP SDK (both majors are optional peers, the project has the one it uses), and `>=0.11.2` on v2 — earlier versions rejected high-level v2 servers in a check `instrument()` swallows, so the integration looked healthy and captured nothing |
| Path B `identify` | Read headers off `extra` directly — exactly the failure the docs PR targeted. v2 puts the request at `extra.http.req` as a WHATWG `Request`, so a v1-shaped read returns `undefined`, `identify()` returns `null`, and **every event goes out anonymous with no error anywhere**. Now uses the exported `getRequestHeaders` |
| Sessions | Nothing on `2026-07-28`. Now notes it removed the handshake and session header, so every request is its own `$session_id` unless `enableConversationId` is set |

## Bundling `sdk-v2.md`

`docs_urls` is a build-time allowlist, so the new page had to be added explicitly — the seven *corrected* pages come along free on the next build, since they're the same URLs.

It goes to **both** variants, not the TypeScript one alone: `description.md` is shared, so a TS-only bundle left the Python skill with five pointers to a file it doesn't ship. The page's own "Python soon, no `2026-07-28` yet" callout is useful there regardless.

Depth stays in the doc; the skill carries only what it needs to stop being wrong.

## Checked

Cold build (`rm -rf .docs-cache`) succeeds; `sdk-v2.md` is bundled and named in `{references}` in both skills; the four corrected pages fetched post-deploy copies. `npm test` 143/143.

⚠️ **Whoever cuts the release**: `DOC_CACHE_TTL_MS` defaults to 24h, so a machine that built in the last day will silently reuse pre-deploy copies of the corrected pages. Build with `DOCS_CACHE_TTL_MS=0` or after `rm -rf .docs-cache`.

No wizard release needed — skills are fetched from the context-mill release at runtime.

## Follow-up: env var mismatch — resolved in posthog.com#19598

The original note here said `sdk-v2.md` was the outlier using `POSTHOG_PROJECT_API_KEY`. Investigation showed the premise was inverted: **every** mcp-analytics docs page used `POSTHOG_PROJECT_API_KEY` (`installation.md` ×6, `start-here.md`, `custom-servers.md`, `sdk-v2.md`, plus the tutorial), while this skill's `description.md` — and the posthog.com-wide and context-mill-wide convention — is `POSTHOG_PROJECT_TOKEN`. So a `sdk-v2.md`-only fix would have left the same skill/docs mismatch via `installation.md`.

PostHog/posthog.com#19598 renames the env var to `POSTHOG_PROJECT_TOKEN` across all five mcp-analytics files. No change needed in this PR — the skill already uses the right name.

⚠️ Same cache caveat as above, one more reason to build fresh: the corrected env var only lands in the bundle if the release is built after the posthog.com PR merges and deploys, with `DOCS_CACHE_TTL_MS=0` or a cleared `.docs-cache`.

---

### DIFF
--- a/context/skills/mcp-analytics/config.yaml
+++ b/context/skills/mcp-analytics/config.yaml
@@ -15,6 +15,8 @@ variants:
     tags: [javascript_node, mcp]
     docs_urls:
       - https://posthog.com/docs/mcp-analytics/installation.md
+      - url: https://posthog.com/docs/mcp-analytics/sdk-v2.md
+        title: MCP TypeScript SDK v1 vs v2 — read when the project uses @modelcontextprotocol/server
       - https://posthog.com/docs/mcp-analytics/custom-servers.md
       - https://posthog.com/docs/mcp-analytics/intent.md
       - https://posthog.com/docs/mcp-analytics/identifying-users.md
@@ -26,6 +28,11 @@ variants:
     tags: [python, mcp]
     docs_urls:
       - https://posthog.com/docs/mcp-analytics/installation.md
+      # Bundled here too, even though it's a TypeScript page: description.md is
+      # shared by both variants and points at it, so omitting it would leave a
+      # dangling reference. Its own callout tells a Python reader what applies.
+      - url: https://posthog.com/docs/mcp-analytics/sdk-v2.md
+        title: MCP TypeScript SDK v1 vs v2 — read when the project uses @modelcontextprotocol/server
       - https://posthog.com/docs/mcp-analytics/custom-servers.md
       - https://posthog.com/docs/mcp-analytics/intent.md
       - https://posthog.com/docs/mcp-analytics/identifying-users.md
--- a/context/skills/mcp-analytics/description.md
+++ b/context/skills/mcp-analytics/description.md
@@ -33,13 +33,16 @@ Follow these steps IN ORDER. Each step has a **TypeScript / JavaScript** part an
 Determine the language first, then route to the matching instructions throughout:
 
 - **TypeScript / JavaScript** — there's a `package.json`. Look for MCP signals in dependencies and source:
-  - `@modelcontextprotocol/sdk` — the official SDK (most common).
+  - `@modelcontextprotocol/sdk` — the official SDK, **v1** (most common).
+  - `@modelcontextprotocol/server` / `@modelcontextprotocol/core` / `@modelcontextprotocol/client` — the official SDK, **v2**. A v2 project has no `@modelcontextprotocol/sdk` at all, so never read that one package's absence as "no MCP server here".
   - `mcp-handler` — the Next.js / Vercel adapter.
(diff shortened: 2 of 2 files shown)


gewenyu99 (commented):
This is great, I'd like to propose a tweak. What concerns me about this PR is the guidance of v1 vs v2 instructions being sprinkled through out instead of having a clear structure/split.

For the class of models we tend to work with in the Wizard, instruction like this is weakly followed, especially if we're 200K context deep. They're less good at round about back and forth reasoning.

What I'd say is under each step to have explicit headings for V1 and V2 with a reminder of the set of instructions to follow. 

so for example:

```ts
### STEP 2: Choose the instrumentation path
shared stuff...

#### For @modelcontextprotocol v1
... stuff

#### For @modelcontextprotocol v2
... stuff
```

The weaker reasoning of these models need a tad bit more hand holding. It'll also speed up these tasks.

We also want to avoid things like:

```ts
// v1: import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js"
// v2: import { McpServer } from "@modelcontextprotocol/server"
import { instrument } from "@posthog/mcp"
```
Comments tend to be ignored (and at random) by models.

gewenyu99 (approved):

gesh:
Restructured per the review feedback in 9a98e73: STEP 2, STEP 3, and STEP 4 now each have explicit `For @modelcontextprotocol/sdk (v1)` / `For @modelcontextprotocol/server (v2)` headings with self-contained instructions, and STEP 1 tells the agent to record the major and follow only the matching sections. The mixed `// v1: ... / // v2: ...` import comments are gone — Path A now has two complete per-major code blocks, and the v2-only notes (`>=0.11.2` floor, the `instrument(server.server)` fix, sessions on `2026-07-28`) live under the v2 headings.

PR #345, opened 2026-08-19
s3 mirror release and smol version check

edwinyjlim:
- workflows to upload release zips to S3 bucket
- small version safeguard that only allows increasing versions

### DIFF
--- a/.github/workflows/mirror-dryrun.yml
+++ b/.github/workflows/mirror-dryrun.yml
@@ -1,0 +1,95 @@
+name: Mirror dry run
+
+# Proves the S3 mirror path end to end without cutting a release.
+#
+# build-release.yml cannot be used for this: its build-and-release job is gated
+# on `is-new-version == 'true'`, which compares package.json to the latest
+# published release. With no pending version bump a workflow_dispatch skips the
+# job entirely, so the mirror steps never run.
+#
+# The OIDC role assumption is the one thing that can only be proven from CI —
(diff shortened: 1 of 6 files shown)


.claude/worktrees/design-doc-context-flow
gewenyu99, line 1: what this

.github/workflows/mirror-dryrun.yml
```diff
@@ -0,0 +1,95 @@
+name: Mirror dry run
```
gewenyu99, line 1: Are these dry runs something we're doing now for testing or keeping longterm? I assume we'll get rid of this entirely once we know this works 👁️

gewenyu99 (commented):
Before I stamp, I have a question. What are the mechanics of mirror-dist?

I understand there's a few differences -> Github release url references vs aws references + we need to rebuild the zip.

What else did we need to handle? Is flattening a mechanical limitation of AWS? Should we flatten both releases for simplicity?

🤔 Would it make sense to find a way for the mirror to just be 1 copy paste and 1 giant search and replace for URL bases?

.claude/worktrees/design-doc-context-flow
edwinyjlim: doh

.github/workflows/mirror-dryrun.yml
edwinyjlim: just to test, but i think your find and replace approach is better. so this might dissapear

gewenyu99 (approved):

PR #348, opened 2026-08-20
fix(skills): scanner skeletons pin a model posthog no longer accepts

TueHaulund:
## Problem

Every scanner the `replay-vision` command (and self-driving step 6c) tries to create fails with a 400: the skeletons pin `"model": "gemini-3.6-flash"`, which posthog retired and remapped to 3.7 (migration 0076). The step-skills treat the 400 as a per-scanner follow-up, so a run "succeeds" while creating zero scanners.

## Changes

- Swap `gemini-3.6-flash` → `gemini-3-flash-preview` in the three replay-vision scanner skeletons and self-driving's 6c reference.
- `gemini-3-flash-preview` is posthog's current default flash tier (5 credits/observation vs 15 for `gemini-3.7-flash`), matching the in-product sizing skill's "start at the default" guidance - it also cuts the skeletons' projected spend to a third.

`node scripts/build.js` passes.

### DIFF
--- a/context/skills/replay-vision-scanner-broken-experiences/description.md
+++ b/context/skills/replay-vision-scanner-broken-experiences/description.md
@@ -20,7 +20,7 @@ Create with `vision-scanners-create`. Two blanks: the `query` and
     ]
   },
   "sampling_rate": 0.5,
-  "model": "gemini-3.6-flash"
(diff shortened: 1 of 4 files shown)


TueHaulund:
Superseded by #349, which carries the same model fix for all four files.

PR #349, opened 2026-08-20
feat(replay-vision): customize the three scanners per app instead of adding more

TueHaulund:
## Problem

`wizard replay-vision` creates three fixed-name, fixed-prompt scanners and only customizes their query scope - but the command's pitch is scanners tailored to the customer's codebase. A booking product gets a scanner literally named "Broken experiences" with a generic prompt. Self-driving's step 6c carries its own drifting copies of the same prompts, and both flows pin `gemini-3.6-flash`, which posthog retired - every create 400s and a run reports success with zero scanners.

## Changes

- Each scanner becomes a **brief**: a locked prompt scaffold (the "unambiguous on screen" bar, what to report, type, sampling, model) with agent-filled blanks - the `name` in the product's own words, the `query` where the brief has one, `{{WATCH_FOR}}` / `{{STUCK_MOMENTS}}` / `{{VOCABULARY}}`, and `{{PRODUCT_CONTEXT}}`. Same three roles, same cost envelope as before; no extra scanners and no `wizard_ask`.
- Shared rules live once in `replay-vision-scanners-core`: naming, the two monitors' query disjointness, untrusted-repo-text handling (blanks are noun phrases, never instruction-shaped; unfillable blanks drop their scaffold sentence rather than invent facts), sizing, and endpoint fallbacks.
- Re-runs match by `scanner_type` + a literal per-brief match phrase + the flow's `emits_signals` value (`false` for `replay-vision`, `true` for self-driving) instead of fixed names. Matches update in place with fresh blanks including the name - legacy fixed-name scanners upgrade on their next run - and `enabled` is preserved so setup never re-arms a scanner a person paused. A name collision gets one rename attempt, then a recorded follow-up.
- Self-driving's step 6c now installs the same two monitor briefs with `emits_signals: true` instead of its own skeleton copies - its footprint is unchanged (two emitting monitors), just customized per product. The summarizer brief stays exclusive to the `replay-vision` command: its unscoped sample overlapping the monitors would self-corroborate findings into promoted reports.
- Briefs pin `gemini-3-flash-preview` (the current default tier, 5 credits/observation). Supersedes #348.

Not included, deliberate: per-brief `credit_limit` values (a product call - the sizing gate covers budget today), and moving the shared repo-profile derivation into the seed handoff so the three parallel scanner tasks stop re-reading the same files (a flow-architecture change for the owning team).

`node scripts/build.js` passes.


edwinyjlim (approved):
look reasonable but haven't tested myself

PR #267, opened 2026-07-22
feat(integration): make Go a full example-based variant

sarahxsanders:
Promotes the Go integration from a **docs-only** variant to a **full example-based** variant, anchored on a real, compiling example app at `example-apps/go`.

**Why:** without an anchoring example, the wizard reinvents the integration shape every run. A worked example collapses that variance — the point raised in review.

**What's here:**
- `example-apps/go/` — idiomatic `net/http` service: one client per process, env-var config, identify, event capture, feature flags via the current `EvaluateFlags` snapshot API, error tracking, and `client.Close()` on graceful shutdown. `go build` + `go vet` clean.
- `config.yaml`: go switched from `template: description-go-docs-only.md` → `example_paths` + `framework: go`.
- Removed the now-unused `description-go-docs-only.md`.

Ships as a pair with the wizard detection PR **PostHog/wizard#945**.

PR #352, opened 2026-08-20
fix(go): name the Identify type so it isn't hand-rolled as a Capture

gewenyu99:
Stacked on #267. The Go commandments named `$set` but no identify API, so a real run built identify out of `posthog.Capture` with a literal `$identify` event name.


gewenyu99:
Wrong fix. The Go docs never teach a server-side `posthog.Identify` call — backend events carry identity via `DistinctId` on the capture, which the existing commandment already says. Closing rather than reworking.

PR #371, opened 2026-08-31
feat(self-driving): drop the in-run GitHub gate

gewenyu99:
Removes step 3 (Connect GitHub) — the wizard now gates on the GitHub App connection in a screen before the agent starts.

Step 3 asked the user to install the GitHub App mid-run via `wizard_ask` and aborted with `[ABORT] github connection declined` if they didn't. Because a `wizard_ask` timeout resolves to the same `__cancelled__` value as an explicit decline, users who simply stepped away had their run killed as if they'd refused. The gate also sat behind a full agent startup, so a hard requirement was discovered late and cost tokens to reach. Moving it to a pre-agent screen makes it a cheap, unambiguous, un-timeoutable check. Step 2 now chains straight to 3b, and `5a-github` / `5b-linear` / `5-connected-tools` / `description.md` are repointed off the removed step. Pairs with the wizard PR that adds the screen — ship this first.

---


edwinyjlim (approved):
lgtm but tagging @PostHog/team-self-driving for a ✅

gewenyu99:
They gave a 👍 in slack so this is goooing in

PR #375, opened 2026-09-01
feat: add google-adk variant to ai-observability skill

marco-g-pm:
## Problem

A Node app built on Google's Agent Development Kit (`@google/adk`) asking the wizard for AI observability has nowhere to land: the framework rules never mention ADK, so the agent falls through to a provider variant and loses the agent structure. PostHog/posthog-js#4687 adds `PostHogADKPlugin`, making the plugin the supported path.

for https://github.com/PostHog/posthog-js/pull/4687

## Changes

- `@google/adk` in the manifest now routes to the new `google-adk` variant (framework table in `1-begin.md`).
- Adds the `google-adk` variant to `config.yaml`, pointing at the Google ADK install doc and the Node library doc.
- Mechanical: variant count 68 to 69 in two files.

## Testing

- All 173 skill tests pass locally with the variant present.
- Not run: `pnpm build`. Doc fetches fail behind this sandbox's TLS interception, and the variant's install doc URL only goes live with the posthog.com page.

## Release notes

Merge wants `mcp-publish` plus a semver label so the variant reaches users. The posthog onboarding PR and the posthog.com page must land first: the build fetches the install doc and fails on a 404.

## 🤖 Agent context

Written by Claude Code from PostHog/posthog-js#4687, mirroring #373 (the Go OTel variant rollout). Skills invoked: writing-pr-descriptions, writing-user-facing-copy.

### DIFF
--- a/context/skills/ai-observability/config.yaml
+++ b/context/skills/ai-observability/config.yaml
@@ -426,6 +426,15 @@ variants:
       - https://posthog.com/docs/ai-observability/installation/mastra.md
       - https://posthog.com/docs/libraries/node.md
 
+  - id: google-adk
+    framework: google-adk
+    display_name: Google ADK
+    description: PostHog AI Observability for the Google Agent Development Kit (agent framework, Node)
+    tags: [google-adk, javascript_node, ai_observability]
+    docs_urls:
+      - https://posthog.com/docs/ai-observability/installation/google-adk.md
+      - https://posthog.com/docs/libraries/node.md
+
   - id: instructor-python
     framework: instructor
     default: true
--- a/context/skills/ai-observability/description.md
+++ b/context/skills/ai-observability/description.md
@@ -6,7 +6,7 @@ Wire up PostHog's AI Observability so calls made through {display_name} land in 
 
 This skill instruments the LLM calls the project *already makes*. It does **not** install the vendor SDK for you.
 
-Check the project's manifest for an LLM package. The catalog is far wider than the obvious providers — 68 variants covering agent frameworks (`openai-agents`, `claude-agent-sdk`, LangGraph, CrewAI, Mastra, …) and OpenAI-compatible gateways (Groq, OpenRouter, Together, Ollama, …), which an app reaches through the `openai` package plus a `baseURL` override. `1-begin.md` carries the ordered decision rules; follow them rather than matching on the first familiar package name. If no LLM SDK is present, switch to the `manual-capture` variant — it posts `$ai_generation` events directly and works standalone.
+Check the project's manifest for an LLM package. The catalog is far wider than the obvious providers — 69 variants covering agent frameworks (`openai-agents`, `claude-agent-sdk`, LangGraph, CrewAI, Mastra, …) and OpenAI-compatible gateways (Groq, OpenRouter, Together, Ollama, …), which an app reaches through the `openai` package plus a `baseURL` override. `1-begin.md` carries the ordered decision rules; follow them rather than matching on the first familiar package name. If no LLM SDK is present, switch to the `manual-capture` variant — it posts `$ai_generation` events directly and works standalone.
 
 Everything else this skill needs — PostHog credentials, instrumentation packages, env vars — the skill installs and configures itself. It does **not** require a pre-existing `posthog.init(...)`. If one is already there, reuse its env-var names in `3-instrument.md`; if not, that step sets fresh values via `set_env_values`.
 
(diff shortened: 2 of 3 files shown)


marco-g-pm:
CI will break until https://github.com/PostHog/posthog/pull/92781 is merged in

context/skills/ai-observability/config.yaml
```diff
@@ -426,6 +426,15 @@ variants:
       - https://posthog.com/docs/ai-observability/installation/mastra.md
       - https://posthog.com/docs/libraries/node.md
 
+  - id: google-adk
+    framework: google-adk
+    display_name: Google ADK
+    description: PostHog AI Observability for the Google Agent Development Kit (agent framework, Node)
+    tags: [google-adk, javascript_node, ai_observability]
+    docs_urls:
+      - https://posthog.com/docs/ai-observability/installation/google-adk.md
```
carlos-marchal-ph: This page 404s today and the build hard-fails on it: `fetchDocOnce` marks 404 non-retryable and `fetchDoc` rethrows when there is no cache, so the `build` check on this PR is already red with `[FATAL] Build failed: ... HTTP 404 Not Found`.

The PR body names the wrong unblocker. PostHog/posthog#92781 adds `docs/onboarding/ai-observability/google-adk.tsx` (the in-app onboarding wizard in the posthog repo). Installation pages live in posthog.com at `contents/docs/ai-observability/installation/*.mdx`, which has no `google-adk.mdx` and no open PR creating one. #373 could build because it reused the existing `installation/opentelemetry.md` rather than adding a page.

context/skills/ai-observability/config.yaml
```diff
@@ -426,6 +426,15 @@ variants:
       - https://posthog.com/docs/ai-observability/installation/mastra.md
       - https://posthog.com/docs/libraries/node.md
 
+  - id: google-adk
+    framework: google-adk
+    display_name: Google ADK
```
carlos-marchal-ph: Every sibling whose framework exists in two languages puts the language in `display_name` (`Google Gemini (Node)`, `Instructor (Node)`, `LangGraph (Node)`). Google ADK has a Python SDK too, so bare `Google ADK` is ambiguous in a way `mastra`/`convex` are not.

This is the label `load_skill_menu` shows and the one `description.md` interpolates into the title, so it is what an agent matches on first. The `description:` line below says Node, but it is read second.

context/skills/ai-observability/references/1-begin.md
```diff
@@ -32,6 +32,7 @@ Apply these rules in order. The first match wins. Frameworks wrap providers, and
 | `instructor` | `instructor-{python,node}` |
 | `litellm` | `litellm` |
 | `mastra`, `@mastra/core` | `mastra` |
+| `@google/adk` | `google-adk` |
```
carlos-marchal-ph: This row only names `@google/adk`, but the variant id it routes to is `google-adk` — the exact PyPI package name for Google's Python ADK (v2.8.0, the more widely used one). A Python ADK project has `google-adk` in its `requirements.txt`, sees a menu variant called `google-adk`, and has an exact-string reason to pick it. This file then reinforces it: "Framework variants have no language suffix" and "Language follows the manifest."

That app would get `libraries/node.md` and `@posthog/ai`, a Node package. Confirmed there is no Python path to fall back to — `posthog-python` ships no ADK adapter — so a guard is the fix, not a new variant. Suggest making the row explicit (Node only; PyPI `google-adk` has no variant) rather than renaming the id. Where Python ADK apps should land instead is your call.

Note the inversion vs `openai-agents`: there the bare id matches the supported package, here it matches the unsupported one.

context/skills/ai-observability/description.md
```diff
@@ -6,7 +6,7 @@ Wire up PostHog's AI Observability so calls made through {display_name} land in
...
+Check the project's manifest for an LLM package. The catalog is far wider than the obvious providers — 69 variants covering agent frameworks (`openai-agents`, `claude-agent-sdk`, LangGraph, CrewAI, Mastra, …) and OpenAI-compatible gateways (Groq, OpenRouter, Together, Ollama, …), which an app reaches through the `openai` package plus a `baseURL` override. `1-begin.md` carries the ordered decision rules; follow them rather than matching on the first familiar package name. If no LLM SDK is present, switch to the `manual-capture` variant — it posts `$ai_generation` events directly and works standalone.
```
carlos-marchal-ph: #373 bumps this same string (and the one in `1-begin.md`) from 68 to 69 as well, so whichever of your two PRs merges second leaves both files claiming 69 when the count is 70.

Nothing in `scripts/lib/tests/` asserts this prose count against `config.yaml`, so the drift ships silently to every agent reading the skill. Either sequence the two merges deliberately or derive the count in a test.

Radu-Raicea (approved):

PR #378, opened 2026-09-01
feat(feature-flags): add wizard feature-flags install skill

fristovic:
The SDK can already evaluate flags. This skill does not add a flags engine. It installs the Next.js App Router pattern the SDK allows and a typical install almost never does: cheap, fast, safe to merge.

Default `wizard` is product analytics. `wizard audit feature-flags` is read-only. MCP / PostHog AI create and manage flags. Neither wires this pattern into the app. This is that install.

**What it puts in the repo (vs SDK defaults)**

- **One `/flags` call per request**, not one per check. `evaluateFlags(distinctId)` once in the root layout, then `flags.isEnabled(key)`. `getFeatureFlag` / `isFeatureEnabled` / `getAllFlags` each hit `/flags` again. PostHog bills those.
- **no second client fetch.** Server values are bootstrapped into `PostHogProvider`. `instrumentation-client.ts` cannot take per-request bootstrap, so existing init there is moved into the provider.
- **Same `distinct_id` on server and client** so percentage rollout is real. Anonymous apps get a `ph_distinct_id` cookie; identified apps reuse the real user id.
- **CI does not poll.** `advanced_disable_feature_flags` in test/CI. No local-evaluation 30s poll on idle servers.
- **Optional worked example, off until they turn it on.** One `wizard_ask`, skip first. Confirm → one boolean flag at **0%** and one additive UI path. Auth / checkout / mutations are out of bounds. The SDK will not stop you from wrapping checkout at 100%.

**Who this is for**

A Next App Router team that wants the cheap/correct first hour, not “can I use flags?”. Skip is valid. If they already evaluate once on the server and bootstrap, they already have the benefit.

Pairs with wizard: PostHog/wizard#1192.

---

Docs: https://github.com/PostHog/posthog.com/pull/19911 (`/docs/feature-flags/installation/ai-wizard`). Do not add that URL to mill `shared_docs` until `https://posthog.com/docs/feature-flags/installation/ai-wizard.md` returns 200.

### DIFF
--- a/context/skills/feature-flags-setup/description.md
+++ b/context/skills/feature-flags-setup/description.md
@@ -1,84 +1,85 @@
-# Add PostHog feature flags (Next.js App Router)
+# Add PostHog feature flags
 
-Use this skill to give a Next.js App Router app a **correct, cheap** feature-flag install: server-side evaluation bootstrapped into the client, one real flag in the user's PostHog project, one gated UI path, and CI configured so forgotten test environments don't poll `/flags`.
+Use this skill to add PostHog feature flags to a **Next.js App Router** app. Once installed, flags are evaluated once per request on the server with `posthog-node`'s `evaluateFlags()`, those values are bootstrapped into the client so the first paint has no flicker and no extra `/flags` fetch, one real boolean flag exists in the user's project, and one UI path is gated behind it.
 
-This is **not** `wizard audit feature-flags` (read-only, after the fact) and **not** the default `wizard` install (product analytics; its prompt explicitly excludes the feature-flags category). This is the missing expert first hour.
+This is **not** `wizard audit feature-flags` (read-only, after the fact). This is **not** the default `wizard` install (product analytics). This skill is the flags install: instrument, create, gate.
 
-## Scope — one stack, one pattern
+## Scope and guardrails
 
-**Supported: Next.js App Router only** (`app/` directory, `next` in `package.json`).
+- **Next.js App Router only.** Require `next` in `package.json` and an `app/` directory (or `src/app/`). If the project is Pages Router, a different framework, or backend-only, **stop**: emit `[ABORT] unsupported stack for feature flags` on its own line and do nothing else. Do not invent a second pattern. Do not add local evaluation. Do not install `@posthog/next`.
+- **One evaluation per request.** Call `evaluateFlags(distinctId)` once, then read with `flags.isEnabled(key)` / `flags.getFlag(key)`. Do not call the deprecated `getFeatureFlag`, `isFeatureEnabled`, or `getAllFlags` — each of those is its own `/flags` request.
+- **Same distinct_id on server and client.** Percentage rollout is deterministic per id. If they differ, bootstrap lies.
+- **Additive gating only.** Flag off = current behavior. Never gate auth, checkout, payments, data-mutation handlers, or middleware that can 404 a route.
+- **Minimal, additive changes.** Match the project's folder style. Read a file immediately before editing it. Do not restructure unrelated code. Do not commit.
 
-If the project is Pages Router, a different framework, or a backend-only package, **stop**: emit `[ABORT] unsupported stack — currently Next.js App Router only` on its own line and do nothing else. Do not invent a second pattern. Do not "also support" middleware-only, local evaluation, or `@posthog/next` (pre-release).
+### Abort cases
 
-**The pattern (do this, nothing else):**
+If anything blocks the run, **always** emit exactly one `[ABORT] <reason>` line and stop — never halt, finish, or error out silently. The wizard catches `[ABORT]` and terminates the run for you; don't try to exit yourself. A silent stop is recorded as a failed run with no reason, which can't be acted on, so every dead end must carry a reason. Use one of:
 
-1. Evaluate flags **once per request** on the server with `posthog-node`'s `evaluateFlags()`.
-2. Pass those values + the same `distinct_id` into the client via `bootstrap` on `PostHogProvider`.
-3. Gate UI with `useFeatureFlagEnabled` from `@posthog/react`.
-4. Create **one** boolean flag at 100% rollout and gate **one** additive UI path after the user confirms.
-
-This avoids flicker (the client has values on first paint) and a duplicate `/flags` request on init. It also avoids local evaluation, whose default 30s poll costs an idle server ~864k requests/month (see `cutting-costs.md`).
-
-## Abort cases
-
-If anything blocks the run, **always** emit exactly one `[ABORT] <reason>` line and stop. The wizard catches `[ABORT]` and terminates the run; don't try to exit yourself. Use one of:
-
-- `[ABORT] unsupported stack — currently Next.js App Router only` — no `app/` directory, or `next` is not a dependency.
-- `[ABORT] could not locate a UI surface to gate` — exhaustive search found no page or component safe to add an additive, flag-gated element to.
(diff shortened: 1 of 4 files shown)


context/skills/feature-flags-setup/description.md
```diff
@@ -0,0 +1,290 @@
...
+    >
+      {children}
+    </PostHogProvider>
+  )
+}
+```
+
+`advanced_disable_feature_flags: true` in test/CI stops forgotten CI jobs from polling `/flags` (see `cutting-costs.md`). A missing token in production is a no-op (render `children`); in development throw the missing-config error named in the framework guidelines.
+
+**If `instrumentation-client.ts` (or `.js`) already inits `posthog-js`:** move that init into this provider so bootstrap can be passed per request. Keep the existing `api_host` / `defaults` / other options. Do not leave both inits in place. Do not follow any framework note that says to keep init in `instrumentation-client.ts` — that path cannot take per-request bootstrap.
```
sarahxsanders: this might work better as a commandment (check and see if it already is, it might be) and see what this might contradict with in commandments in general

context/skills/feature-flags-setup/description.md
```diff
@@ -0,0 +1,290 @@
...
+
+Emit `[STATUS] Detecting Next.js App Router`.
+
+Look for `next` in `package.json` **and** an `app/` directory (or `src/app/`). The lockfile decides the package manager (`pnpm-lock.yaml`, `package-lock.json`, `yarn.lock`, `bun.lockb`).
+
+If this is not Next.js App Router, apply the unsupported-stack abort. In the same message, say where you looked.
+
+Record: package manager, whether `src/` is used, and whether PostHog is already initialized (`posthog.init`, `PostHogProvider`, `instrumentation-client`, `posthog-js` / `posthog-node` in dependencies). If flags are already wired the way this skill describes (`evaluateFlags` + bootstrap + a gated call site), verify they are correct and skip to STEP 10.
+
+### STEP 2: Credentials
```
gewenyu99: If we're installing the package, we can init this for them. There's examples you can copy from, the OAuth session will let you grab the token on the wizard side to add them. I think it'd be pretty odd to have an environment variable -> have no integration.

context/skills/feature-flags-setup/description.md
```diff
@@ -0,0 +1,290 @@
...
+### STEP 4: Server client
+
+Emit `[STATUS] Adding the server client`.
+
+Add a small helper, matching the project's folder style (`lib/posthog-server.ts` or `app/posthog.ts`):
+
+```ts
+import { PostHog } from 'posthog-node'
+
+export function PostHogServer(token: string) {
```
gewenyu99, line 90: I'm gonna throw you one more wrench, you'll need to figure out a reasonable lower bound for the patterns you apply here. The way Next versions work across the board differs, and changes how PostHog need to be initialized. There are examples here in the repo

context/skills/feature-flags-setup/description.md
```diff
@@ -0,0 +1,290 @@
...
+
+Do not introduce local evaluation.
+
+### STEP 6: Confirm the gate target
+
+Emit `[STATUS] Asking which UI path to gate`.
+
+Flags change production UI. Ask **once**, then create a flag and gate only if they picked a target.
+
+1. Scan pages and components for additive surfaces (a banner, an extra card, an empty-state illustration). Prefer a new element over wrapping existing critical logic.
```
gewenyu99, line 142: 🤔 Do you think someone is more likely to come to this program with a thing they have in mind and wanna gate, or do you think they'd be looking for suggestions on what to gate?

context/skills/feature-flags-setup/description.md
```diff
@@ -0,0 +1,290 @@
...
+
+The root layout is a Server Component. If `NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN` is missing, render `{children}` without evaluating or wrapping — boot must still work. Otherwise evaluate once and pass the snapshot into the provider.
+
+Only read a specific key when STEP 7 created (or reused) one. Do not invent a key to satisfy this snippet. Bootstrap docs drop false and empty values — at 0% rollout `getFlag` is `false`, so `featureFlags` is `{}`. That is correct: the client matches "off."
+
+```ts
+const token = process.env.NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN
+const cookieStore = await cookies()
+const distinctId =
+  cookieStore.get('ph_distinct_id')?.value ?? crypto.randomUUID()
```
gewenyu99: PostHog _always_ generates a distinct_id on the frontend. You can call `posthog.get_distinct_id()`. If they're not identified, they'd have an anonymous distinct Id that's merged with the user on identify.

gewenyu99 (commented):
Hey, this is a great start. I have some suggestions:
- I'm not sure if I understand who would want/need this Wizard
- Try setting up feature flags yourself, think about the pain points/difficult bits that someone is likely to fumble on, see if this resolves those issues.
- I would think about this program's relationship to the main PostHog integration program and shared skill reuse. I think this contains a less reliable but duplicate version of many of they paths.

context/skills/feature-flags-setup/description.md
fristovic: It seems like this already is a commandment, I missed that. I'll have to add an exception here I think because we want this skill to do the opposite so it can pass per-request `bootstrap`.

I will add a `nextjs-flags-bootstrap` commandment and tagg only this skill.

context/skills/feature-flags-setup/description.md
fristovic: Both can be right depending on the user. That's why I built both options here:

- Skip first = "I already know what I'll gate" (they add `useFeatureFlagEnabled` on the wiring we just installed)
- Suggestions = "show me the loop, I don't know how this works"

Right? Feel free to correct me If I'm missing something here?

context/skills/feature-flags-setup/description.md
fristovic: Actually, thats not a wrench at all, makes total sense. I went with Next.js App Router 15.3+, matching the mill Next commandment and `example-apps/next-app-router` (15.5.x). 
We read `package.json`'s `next` version and abort below that. On 15.3+ we still relocate `instrumentation-client` into the provider when bootstrap is required.

context/skills/feature-flags-setup/description.md
fristovic: Yep, this was a sequencing leftover. This skill will now require an existing client init. OAuth still fills a missing `phc_` into existing env names. If there's no init, we abort and point at `npx @posthog/wizard` instead of shipping a second integration.

fristovic:
> Hey, this is a great start. I have some suggestions:
> 
> * I'm not sure if I understand who would want/need this Wizard

I can see this easily being used by someone who wants to evaluate cheaply feature-flags (server eval, bootstrap, CI off). I don't see it as a flag factory. Default wizard won’t install this pattern.

On the fumbles: the ones I initially meant to hit are double `/flags`, first-paint render issue, CI polling overnight, and the deprecated per-call APIs (`getFeatureFlag` / `getAllFlags`). The weaker copy of the main integration (init, env, packages, Next version) was a real miss that I will address in the other comments.

On reuse: agreed. This skill will now extend an existing PostHog install instead of reimplementing it.

fristovic:
Official docs page (draft): https://github.com/PostHog/posthog.com/pull/19911 — `/docs/feature-flags/installation/ai-wizard`. Do not add that URL to `shared_docs` until it returns HTTP 200.

PR #381, opened 2026-09-02
add /latest path to context mill aws cdn

edwinyjlim:
Making the latest context mill version available in AWS CDN + S3 at `{aws-cdn}/latest/*`


edwinyjlim:
forgive me. yolo.

PR #382, opened 2026-09-03
fix(integration): initialize snippets with read-only toString

marandaneto:
## Problem

The browser snippet embedded in seven integration examples assigns `toString` directly on an array stub. In strict mode, that assignment throws when a site makes `Array.prototype.toString` read-only, so PostHog does not initialize.

This is the context-mill counterpart to PostHog/posthog#94098.

## Changes

- Generate all seven embedded snippets from one source while preserving each example's method list and asset host behavior.
- Define own `toString` properties with writable, configurable, and enumerable descriptors instead of assigning through the inherited array property.
- Preserve the existing `posthog (stub)` and `posthog.people (stub)` output.
- Check that generated copies stay in sync with the source.
- Run every snippet variant in strict mode with read-only `Array.prototype.toString` in the regression test.

## Testing

- `pnpm check:browser-snippets`
- `pnpm vitest run scripts/lib/tests/browser-snippet.test.js`
- `pnpm test`
- `BUILD_VERSION=1.50.0 pnpm build`

The build generated safe integration `EXAMPLE.md` files in `dist/marketplace`.

## Downstream regeneration

Generated PostHog/skills and PostHog/ai-plugin files are not edited in this PR. The supported path runs during a context-mill release: `.github/workflows/build-release.yml` runs `pnpm build`, reads `dist/push-manifest.json`, and mirrors the generated marketplace plugins to PostHog/skills plus the omnibus skills to PostHog/ai-plugin. The next context-mill release will propagate this change automatically.


ioannisj (approved):

package.json
```diff
@@ -29,10 +31,5 @@
     "@posthog/warlock": "^0.2.2",
     "gray-matter": "^4.0.3",
     "js-yaml": "^4.1.1"
-  },
-  "pnpm": {
```
marandaneto, line 33: > Run pnpm check:browser-snippets
 [WARN] The "pnpm" field in package.json is no longer read by pnpm. The following keys were ignored:
 "pnpm.onlyBuiltDependencies". See https://pnpm.io/settings for the new home of each setting.
 $ node scripts/generate-browser-snippets.js --check
 All embedded browser snippets are current.

The warning was harmless because pnpm-workspace.yaml already contains the correct onlyBuiltDependencies setting

PR #384, opened 2026-09-03
feat(flag-health): add PostHog flag roster alignment audit skill

talagluck:
## Summary
- New `flag-health` audit skill (`wizard audit flag-health`) that cross-references a project's feature-flag call sites against PostHog's live flag roster: ghost flag keys, active-but-unreferenced flags, stale 100%/0% rollouts still branched on in code, multivariate flags read as booleans, and long-lived flags with no lifecycle tag.
- Fixes found by dogfooding the skill end-to-end (real PostHog project + several example apps, via parallel subagents running each check for real):
  - Presence-detector grep now requires call-site syntax (`\(`) and excludes docs/node_modules/vendor — it previously false-triggered a full 5-subagent audit on 13/40 example apps that had zero real flag usage (matches were the PostHog snippet loader's stub method-name list, SDK config keys, or README prose).
  - `ff-ghost-flag-key`, `ff-stale-full-rollout`, and `ff-multivariate-as-boolean`'s grep patterns now consistently cover snake_case SDKs (Python `feature_enabled`/`get_feature_flag_payload`, etc.) — previously only one of the three tasks special-cased Python, so the other two undercounted or missed call sites in non-JS codebases.
  - `ff-ghost-flag-key`'s "dynamic key → skip" guard now resolves one hop through a local static flag registry (e.g. `flag.key` from an array of `{ key, ... }` objects) before giving up — the previous wording could cause a real ghost flag to be silently skipped as "dynamic" in that common pattern.
  - Report template restyled to match this repo's other audit skills (`audit-feature-flags`): three-sentence recommended-action items (what's wrong / why it matters / how to fix), explicitly defined status glyphs, an Assumptions-and-blind-spots section, an About-this-audit closer, and fix wording that fits `ff-ghost-flag-key`'s case (there's no existing PostHog flag to stage a removal against).

## Test plan
- [x] `npm test` — 173 tests pass
- [x] `npm run build` — skill packages cleanly into `dist/skills/flag-health.zip`
- [x] Ran the skill for real (subagent-driven, PostHog MCP live) against a real app with its own PostHog project, and against a context-mill example app — all 5 checks landed on independently-verified-correct results in both runs

https://claude.ai/code/session_01HBZpxnSyEjzrmLLUpgPVR3

### DIFF
--- a/context/skills/flag-health/config.yaml
+++ b/context/skills/flag-health/config.yaml
@@ -1,0 +1,18 @@
+type: skill
+template: description.md
+description: Audit where a PostHog integration's code and its actual flag configuration have drifted apart — ghost keys, stale rollouts, undocumented long-lived flags, unreferenced active flags, and multivariate flags misread as booleans
+tags: [best-practices]
+cli:
+  role: command
+  parentCommand: audit
+  command: flag-health
+references:
+  preamble: "**Read ONLY this file.** Do not read any other reference file until this one tells you to."
+shared_docs:
+  - https://posthog.com/docs/feature-flags/cutting-costs.md
+  - https://posthog.com/docs/feature-flags/cleaning-up-stale-flags.md
+variants:
+  - id: all
+    display_name: PostHog audit — flag roster alignment
+    tags: [best-practices]
+    docs_urls: []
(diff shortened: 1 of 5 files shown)


talagluck:
Reopening against my own fork instead of upstream — closing this one.

PR #387, opened 2026-09-04
feat(skills): add cull-feature-flags skill

johncwaters:
Skill half of `wizard cull-feature-flags`; the wizard PR (PostHog/wizard#1210) carries the
command and the deterministic scan and classification. Opening as a draft.
The commits are chunked so the steps can be read in order.

## Shape

A three-step docs-only chain with one Next.js variant, keyed by framework the
same way `integration/config.yaml` does it.

1. **Verify.** Read each seeded ledger row at its `file:line`, confirm the
   proposed action or downgrade the row to keep (never the other way round),
   and resolve rows one at a time so the run screen moves.
2. **Cull.** One `wizard_ask` call with two questions: "Report only, change
   nothing" as its own single choice first, then the flag multi-select.
   Approved rows get the code edit first, then `feature-flag-disable` through
   `exec` (search, info, call). The skill never deletes or archives, and skips
   the disable for flags that are already archived or off.
3. **Report.** Findings table first (stale, warning or healthy per flag) so a
   report-only run still gives something to act on, then Culled, Left for you,
   Failed and Kept, plus the undo block.

The skill never classifies anything itself. The ledger the wizard seeds is
ground truth. Areas are the human bucket names, defined once in the
description: Rolled out, Never enabled, Archived in PostHog, Disabled in
PostHog, Unreferenced, Comment only, Dead code, Deleted in PostHog, Many call
sites, Healthy.

Ships as `role: skill` per the promotion criterion in CONTRIBUTING; the wizard
registers the flat command on its side.

## Consent by lane (second round)

Reviewers would not run a flat multi-select of eight flags with pre-checked boxes and 100% kill-switch-looking flags next to 0% rollback levers. Step 2's single `wizard_ask` now asks one pick list per lane, only for lanes with proposals: Rolled out; Off for everyone (with the rollback caveat, keeping costs nothing); Archived, Disabled or Deleted in PostHog; Unreferenced, Comment only or Dead code (says only this repository was scanned). Then the report-only or cull choice last, as the confirm step. Every option's description is that flag's plan: reason, the branch kept in plain words, every call site, and "disables the flag" or "no PostHog change". Nothing is pre-selected. Report only emits its own status line so the wizard shows the outcome. Zero proposals skip the ask entirely. The lane table in the description is shared word for word with the wizard's run screen (PostHog/wizard#1210).

## Testing

`npm test` and `npm run build` are green; the zip carries `SKILL.md` plus the
three references with the preamble and continuation links in place. One
interactive run against a planted Next.js repo
(https://github.com/johncwaters/nextjs-stale-flags, seed it into your own
project per its README): 9 rows verified, one consent
prompt, 8 culled, 0 failed, and the report's findings matched the fixture's
expected verdicts 19/19.

## Running it locally

    pnpm dev                      # in this repo, serves skills on :8765

    cd ../wizard
    pnpm try --local-context-mill --capture-aio --project-id=590630 \
      --install-dir=/path/to/nextjs-app cull-feature-flags


context/skills/cull-feature-flags/references/2-apply.md
```diff
@@ -0,0 +1,73 @@
...
+
+`Read` `.posthog-audit-checks.json`. Collect every row with `status: "warning"` whose `area` is not `Many call sites` (that bucket is report-only).
+
+Zero such rows: skip to the Output section, nothing to apply.
+
+Call `mcp__wizard-tools__wizard_ask` exactly once, with two questions in that one call so the report-only choice stands on its own instead of hiding in the flag list:
+
+1. `id: "mode"`, `kind: "single"`, prompt verbatim:
+
+   > Nothing has been changed yet. Do you want to cull flags now, or just get the report?
```
gewenyu99: Solid choice

context/skills/cull-feature-flags/references/1-verify-call-sites.md
```diff
@@ -0,0 +1,57 @@
...
+
+Emit at the start:
+
+```
+[STATUS] Reading the flag cull ledger
+```
+
+## Read the ledger
+
+`Read` `.posthog-audit-checks.json` at the project root.
```
gewenyu99, line 19: Mmmm creative use of this ledger

gewenyu99 (commented):
On vibes this looks solid

context/skills/cull-feature-flags/references/2-apply.md
```diff
@@ -0,0 +1,73 @@
...
+
+   > Pick the flags to cull. Each one gets disabled in PostHog (re-enable any time from the flag page) and its check removed from code (revert with git).
+
+   One option per row: label `[<area>] <key>: <proposed action from label>`, value `<key>`.
+
+If the call errors (non-interactive host, cap reached), treat it as report-only. Do not retry more than once.
+
+`mode` is `report-only`: nothing is approved, whatever `flags` holds. Otherwise the approved rows are exactly the `flags` answer.
+
+Rows not approved: resolve to `status: "pass"` with `details` = seeded details plus `; declined by user`.
```
edwinyjlim: this is v cool and powerful but it's hard to navigate. we're lacking a lot of UI components to work with, so this isn't a huge priority, but how can we break up the culling selection? 

more importantly, how can we use that moment to explain/teach the developer why this flag should be culled (stale, dead code, unreferenced, etc.)?

Image

sarahxsanders (commented):
this is great!! the guardrails in here are great, looks like you're keeping the model from calls it's genuinely bad at lolol

some tips (and these really are just nits):
- think about what might _look_ like a disabled flag but really isn't, and build in some guardrails there too
- maybe some verification/typechecking/linting for things you want verified

overall, it's nice :) these are really just super fine nits

context/skills/cull-feature-flags/references/2-apply.md
johncwaters: Thanks. The decline row got lost in the multi-select, so mode became its own question.

context/skills/cull-feature-flags/references/1-verify-call-sites.md
johncwaters: On purpose, the ledger already had the tui and the resolve tool. Open to a rename.

context/skills/cull-feature-flags/references/2-apply.md
johncwaters: Agreed. For now each flag carries a one-line why and effect and the list is seeded in bucket order. Testing a few UI options for the pick itself, per-bucket questions being one, before settling on a shape.

johncwaters:
Thanks. The skill now runs lint and typecheck on the touched files before any disable, and failing rows never reach PostHog. The false-disabled guards are a good call, same answer as on the wizard PR: worth doing properly, not today.

context/skills/cull-feature-flags/references/2-apply.md
```diff
@@ -0,0 +1,123 @@
...
+A row whose file still fails resolves to `error` in the format above and never reaches the disable pass.
+
+### Disable and resolve
+
+Before the loop below, run `exec({ "command": "search feature-flag" })` once, pick the tool whose description says it disables a flag, then run `exec({ "command": "info <tool_name>" })` once.
+
+For each approved row whose code edit and file verification passed, in ledger order:
+
+1. `Deleted in PostHog` has no flag, and `Archived in PostHog` and `Disabled in PostHog` are already off: these rows need no PostHog call, go straight to step 4.
+2. Emit `[STATUS] Disabling <key> in PostHog`, then run only `exec({ "command": "call <tool_name> <json> })` with the flag key or id from `details`.
```
gewenyu99, line 107: wait, the local changes haven't been deployed yet? This would break production, no?

context/skills/cull-feature-flags/references/1-verify-call-sites.md
```diff
@@ -0,0 +1,65 @@
...
+
+| bucket (`area`) | confirm as `warning` when | downgrade to `pass` when |
+|---|---|---|
+| `Rolled out` | the call site is a boolean check whose true branch is the current behaviour | the flag gates something that must stay switchable (kill switch, ops toggle named as such) |
+| `Off for everyone` | the call site is a boolean check whose false branch is the current behaviour | the feature is clearly mid-build (recent scaffolding, TODOs pointing at it), or the flag or its call site reads like a kill switch or rollback (name, comment, a recent change back to 0%) |
+| `Archived in PostHog`, `Disabled in PostHog` | the call site is a boolean check; keep the false branch | never, these are always safe to propose |
```
gewenyu99, line 40: > never, these are always safe to propose 
Not always true. Sometimes we temporarily disable a flag. I think premise is correct but wording can be tweaked

context/skills/cull-feature-flags/references/2-apply.md
```diff
@@ -0,0 +1,123 @@
...
+7. `Deleted in PostHog`: code edit only, there is no flag to disable.
+
+A failed code edit resolves the row to `status: "error"` with `details` = the seeded details plus `; failed: <one-line reason>`. That row is finished: it never reaches the disable pass, and PostHog is never touched for it.
+
+### Verify the edited files
+
+Once per run, after every approved row has been edited:
+
+1. Emit `[STATUS] Type checking <n> files`.
+2. `Read` `package.json` for the project's lint and typecheck scripts. Run them only on the files edited in this session, never across the whole project. Capture stdout and stderr; truncate long output to the failure region.
```
gewenyu99, line 95: In this case, this isn't quite true actually. Broken imports can happen cross package, for verification like type checks, definitely run the whole project

context/skills/cull-feature-flags/description.md
```diff
@@ -0,0 +1,62 @@
+# Cull stale PostHog feature flags in a {display_name} project
+
+This skill removes feature flags that have outlived their purpose: rolled out to everyone, off for everyone, archived or deleted in PostHog but still checked in code, or defined in PostHog and never evaluated anywhere. The wizard already did the detection before you started: it scanned the source tree for flag calls, fetched the project's flags, classified every flag with plain rules, and wrote one row per flag into `.posthog-audit-checks.json` at the project root. That ledger is the ground truth. You never grep for flags, never re-classify a row, and never promote a healthy flag into a removal.
```
gewenyu99, line 3: I understand the intent, but disabling greps outright probably:
- really messes with the ability for the agent to properly carry out the task to independently verify. In the wild, flags are weird and projects are weird
- The agent will just use another tool call xD

I think we need to restrict a slightly different way

gewenyu99 (changes requested):
There's definitely a few issue here, especially with the order which we remove flags after disabling them in code. There's a whole deploy process 🤔

johncwaters:
While interesting and fun, there are so many things wrong with this approach that this PR simply needs: culled. :)

PR #388, opened 2026-09-04
fix(integration): capture on confirmed success, and stop over-flagging person processing

sarahxsanders:
Two instrumentation-quality bugs from @raquelmsmith, plus the audit checks that catch them.

1. **Events fired on submit intent, so a form the server rejected still counted as a success.** Fixed: completion events go in the branch that runs after the response confirms success.

2. **Server Action forms had nowhere legal to capture from.** `useActionState` never hands the client a response, and the React rule forbids reacting to state in an effect. Fixed: capture inside the action on the server; client fallback fires off the returned success state.

3. **Backend SDKs overwrote person properties and billed as identified.** Fixed: server events that only record that something happened pass `$process_person_profile: false`.

4. **First version of that rule went too far — it flagged signup and subscription too, leaving no person profiles at all.** Fixed: person-defining events are let through and paired with `identify()`/`$set`.

5. **`wizard audit events` couldn't resolve a single check.** Nothing seeded its ledger, so every id came back unknown. Fixed: step 1 seeds its own.

6. **`audit-identify` only ever penalised a missing flag, never a flag on everything.** Fixed: that check reads both ways now.

Also adds `capture-fires-on-success` to the audits, so people can find intent-fired events in code the wizard already wrote.

## Testing

Four live wizard runs (Next.js, Fastify, Django ×2) via `--local-context-mill`; `pnpm build && pnpm test` green and warlock clean. On the Django app the flag went from 13/13 captures to 3/12 — only the activity events keep it — and `identify()`/`$set` went 0 → 8.

## Checklist

- [x] Skills build, new text reaches all 40 variants
- [x] `pnpm test` passing
- [x] Warlock scan clean
- [x] Verified end-to-end against real apps

Land [PostHog/wizard#1211](https://github.com/PostHog/wizard/pull/1211) first — it seeds the new check id.

---


edwinyjlim:
@sarahxsanders looks solid to me. let's run a few CI jobs beginning of next week

PR #398, opened 2026-09-17
docs(ai-observability): route by generation path, not manifest presence

gewenyu99:
## What
Two selection-rule fixes in the `ai-observability` skill:

- **`1-begin.md` rule 2** — a manifest package no longer auto-selects the framework variant. The framework wins only when it carries the app's generations; if it's a side path (datastore/vector-store provider, retrieval, eval) and a direct provider SDK makes the primary calls, route by that provider.
- **`2-install.md`** — the openai wrapper and every `opentelemetry-instrumentation-*` bridge target `openai>=1.x`. On a pinned pre-1.0 `openai` (`openai.ChatCompletion`/`openai.Embedding`), use `manual-capture`; the wrapper/instrumentor have nothing to patch and silently emit nothing.

## Why
Found while testing the wizard's `ai-observability` program end to end on real OSS repos. On one repo, `llama-index` was in the manifest but only as one optional datastore provider, while the real LLM calls were raw `openai==0.27.5` in a services module. The agent picked the LlamaIndex/OTel path per rule 2, wrapped only the datastore, missed the real calls, and reached for an OTel instrumentor that can't patch the pinned pre-1.0 SDK — likely emitting zero events. Grounding on the actual call sites produces the correct instrumentation; these edits make the rules say so.

---


context/skills/ai-observability/references/2-install.md
```diff
@@ -21,6 +21,8 @@ Every provider and every OpenAI-compatible gateway needs the PostHog SDK next to
 
 The vendor SDK is already in the manifest. Do not add or upgrade it.
 
+**Watch the pinned major.** The openai wrapper and every `opentelemetry-instrumentation-*` bridge target the modern client API (`openai>=1.x`, `client.chat.completions.create`). A project pinned to a pre-1.0 `openai` (the flat `openai.ChatCompletion` / `openai.Embedding` module API) gives them nothing to patch, and you must not upgrade it — so the instrumentation silently emits no events. Use `manual-capture` and wrap the real calls with `posthog.capture()` instead. The same holds for any provider pinned below its current client interface.
```
gewenyu99:
```suggestion
**Watch the pinned major version.** The openai wrapper and every `opentelemetry-instrumentation-*` bridge target the modern client API (`openai>=1.x`, `client.chat.completions.create`). A project pinned to a pre-1.0 `openai` (the flat `openai.ChatCompletion` / `openai.Embedding` module API) gives them nothing to patch, and you must not upgrade it — so the instrumentation silently emits no events. Use `manual-capture` and wrap the real calls with `posthog.capture()` instead. The same holds for any provider pinned below its current client interface.
```

edwinyjlim (approved):

PR #400, opened 2026-09-18
fix(audit): seed the ledger in each narrow audit skill

gewenyu99:
## Problem

Five audit leaves tell the agent that the runtime hands them a ledger:

> The audit ledger is seeded by the wizard with one pending check per … check.

It does not. The wizard pre-seeds only the comprehensive `audit` skill (and the retired `events-audit` program). A skill promoted to an audit-family leaf runs through the wizard's generic agent-skill program, which seeds nothing, so today `wizard audit events` and its four siblings run with **no ledger on disk**:

- every `audit_resolve_checks` call fails with `unknown check id(s)`
- the wizard's "Audit plan" tab stays empty for the whole run
- the report step, which renders directly from the ledger, emits `[label]` placeholders
- the same emptiness reaches the task stream

The same files contradict themselves: each one's check-entry section already says "This skill **seeds** N areas".

## Change

Each skill seeds its own checklist in step 1, the pattern `audit-attribution` already uses — it states outright that the runtime does not pre-seed, and calls `audit_seed_checks` with an explicit payload.

| Skill | Rows seeded | Areas |
|---|---|---|
| `audit-events` | 7 | Event Capture, Event Capture — Optimize |
| `audit-autocapture` | 6 | Autocapture, Autocapture — Optimize |
| `audit-feature-flags` | 9 | Feature Flags, Feature Flags — Optimize |
| `audit-session-replay` | 8 | Session Replay, Session Replay — Optimize |
| `audit-identify` | 15 | Identification, — Lifecycle, — Optimize, — Server SDK |

Every id is one the skill's own steps resolve, in the order the steps reach them, with a label written from that check's rule text. `audit-identify` also declares the server-SDK area its step 5 uses; its check-entry section previously named three areas and omitted it.

The report steps said "Every check the wizard seeded for this skill ends up in the report" — now "Every check Step 1 seeded". That line is corrected in `audit-attribution` too, which already seeded itself.

## Verification

- `node scripts/build.js` succeeds, and the built `audit-events.zip` and `audit-identify.zip` carry the seed call.
- `vitest run scripts/plugins/tests scripts/lib/tests` — 24 files, 207 tests pass.
- Checked every seeded id against the ids each skill's steps resolve: no missing rows, no duplicates, no label over 55 characters, every area inside the wizard's 18-character area column.

Labels and areas are what the wizard renders in its live check list, and after [PostHog/wizard#1263](https://github.com/PostHog/wizard/pull/1263) the areas are also what the task stream publishes as progress rows.

---


edwinyjlim (approved):
man these are kinda messy

